PR #4, opened 2018-04-09
Update README.md with screenshot from Wikimedia.

speps:
This is usually pretty useful in READMEs.


craigwi:
I incorporated this change into some other modifications I made to readme.md.  Thanks!

PR #15, opened 2018-04-09
Fix stack buffer overflow in IDM_STARTCMDSHELL

xiaoyinl:
The buffer `szParams` has the same length as max path length. When you copy the format string ` -Single -Dir "<szDir inserted here>"`, it can overflow `szParams` by 4 bytes if `szDir` has the maximum length (247 bytes). This bug isn't exploitable, but it crashes winfile if I open the cmd shell with a long directory path.

This PR fixes it by increasing the length of `szParams`, and use `wnsprintf`.

### DIFF
--- a/src/wfcomman.c
+++ b/src/wfcomman.c
@@ -961,7 +961,7 @@ AppCommandProc(register DWORD id)
 			   }
 			   lstrcat(szToRun, TEXT("\\ConEmu\\ConEmu64.exe"));
 			   if (PathFileExists(szToRun)) {
-				   wnsprintf(szParams, MAXPATHLEN + 20, TEXT(" -Single -Dir \"%s\""), szDir);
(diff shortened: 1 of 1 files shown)


src/wfcomman.c
```diff
@@ -961,7 +961,7 @@ AppCommandProc(register DWORD id)
 			   }
 			   lstrcat(szToRun, TEXT("\\ConEmu\\ConEmu64.exe"));
 			   if (PathFileExists(szToRun)) {
-				   wsprintf(szParams, TEXT(" -Single -Dir \"%s\""), szDir);
+				   wnsprintf(szParams, MAXPATHLEN + 20, TEXT(" -Single -Dir \"%s\""), szDir);
```
bovine: maybe use `sizeof(szParams) / sizeof(szParams[0])` rather than hardcoding buffer size, which might become desynced over time

xiaoyinl: Fixed. Thanks for the suggestion!

craigwi:
Good catch.  I fixed this one and another similar case.

PR #21, opened 2018-04-09
Restore dogfooding in XP

roytam1:
- Downgrade VC project to VC2010
- Downgrade C/C++ syntax to VC2010
- Dynamic loading Wow64 functions

### DIFF
--- a/src/wfgoto.cpp
+++ b/src/wfgoto.cpp
@@ -470,7 +470,7 @@ vector<PDNODE> *GetDirectoryOptionsFromText(LPTSTR szText, BOOL *pbLimited)
 
 	for (int i = 0; i < options_per_word.size(); i++)
 	{
-		vector<PDNODE> * option = options_per_word.at(i);
+		vector<PDNODE> *option = options_per_word.at(i);
 		// if we didn't return one of the options as is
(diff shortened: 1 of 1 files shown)


src/wfgoto.cpp
```diff
@@ -465,8 +468,9 @@ vector<PDNODE> *GetDirectoryOptionsFromText(LPTSTR szText, BOOL *pbLimited)
 
 	vector<PDNODE> *final_options = TreeIntersection(options_per_word);
 
-	for (auto option : options_per_word)
+	for (int i = 0; i < options_per_word.size(); i++)
 	{
+		vector<PDNODE> * option = options_per_word.at(i);
```
cfrank-rv: Nit: follow code style for pointerAlignment

`vector<PDNODE> *option = options_per_word.at(i);`

craigwi:
I'm open to changes to the .c/.h sources which make this use case possible, but I will not downgrade the vcxproj file.  You will need to create a separate one to compile with anything other than VS2015 and VS2017.

NazmusLabs:
Yeah, I agree with craigwi. We shouldn't downgrade to an older project file.

PR #23, opened 2018-04-09
Fix cmd and notepad hijacking

xiaoyinl:
Pass full path of cmd.exe and notepad.exe to ExecProgram. Otherwise, if the current directory also has a file named cmd.exe or notepad.exe, then it is launched, rather than the one in the system directory.

### DIFF
--- a/src/wfcomman.c
+++ b/src/wfcomman.c
@@ -703,7 +703,12 @@ OpenOrEditSelection(HWND hwndActive, BOOL fEdit)
           }
 
           if (cchEnv == 0)
-            lstrcpy(szToRun, TEXT("notepad.exe"));
+          {
+              if (GetSystemDirectory(szToRun, MAXPATHLEN) > 0)
+                  lstrcat(szToRun, TEXT("\\notepad.exe"));
+              else
+                  lstrcpy(szToRun, TEXT("C:\\Windows\\System32\\notepad.exe"));
+          }
 
           ret = ExecProgram(szToRun, szPath, NULL, (GetKeyState(VK_SHIFT) < 0), FALSE);
       }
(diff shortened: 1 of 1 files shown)


ZanderBrown:
But is this a bug or a feature.....

xiaoyinl:
@ZanderBrown Well, I think it's a bug, because in general users don't expect an executable from the current directory can be launched when they try to open the command shell or edit a txt in notepad. Also, I think it's usually a good idea to pass full path to ShellExecute.

exchange12rocks:
Windows is not always installed on the "C:" drive. If you want to explicitly point to the system32 folder, use appropriate functions to determine its path.

craigwi:
I like the spirit of the change, but it needs to be changed per @exchange12rocks comment.

xiaoyinl:
@exchange12rocks @craigwi You are right. But that's exactly what I am doing. It first calls `GetSystemDirectory` to get the system32 folder. It falls back to a hardcoded C:\Windows\System32, only if the API fails. But I don't think it will ever fail since our buffer is long enough. So I don't think it really matters what to do when it fails. Or do you want me to initialize the buffer to empty string in that case?

sylveon:
You could use `SHGetKnownFolderPath`. With this one you don't have to worry about buffer size.

craigwi:
Fixed this in a slightly different way.  Thanks!

NazmusLabs:
What's for need for all these complicated conditions to determine system32 path. You can use "%windir%\system32" to get to the folder regardless of what drive Windows is on...

sylveon:
And what if someone clever did

set WINDIR=Z:\MyFakeWin
winfile.exe

2018-04-29 22:49 UTC−04:00, NazmusLabs <notifications@github.com>:
> What's for need for all these complicated conditions to determine system32
> path. You can use "%windir%\system32" to get to the folder regardless of
> what drive Windows is on...
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/Microsoft/winfile/pull/23#issuecomment-385304303


-- 
Envoyé de mon Gmail

NazmusLabs:
Fair point. Though I guess if your environment variables is messed up this badly, you probably need to refresh Windows, as a lot of apps use the environment variables and some other program is bound to call Windir. But you can also go further. What if some malicious program replaces cmd.exe with a stub that launches something else. There doesn't seem to be a foolproof wah to launch cmd, but I guess using the get dir function can be seen as more secure than %Windir%, sure.

PR #33, opened 2018-04-09
Replaced Icon

Techokami:
Replaced the venerable old icon with one from the cabview program in Windows 10. This looks much nicer in a Windows 10 environment. The original icon is kept for posterity.

I have no idea if this will be accepted or not, in multiple ways. Here goes nothing! 👀💦


ZanderBrown:
Not sure I like this idea, perhaps departs too much from the original?
Including higher resolutions of the old icon would be good though

NazmusLabs:
There's no need to have multiple ico files. One icon file can hold multiple icons. I would combine the original winfile icon with the Vista based icon in Win10 into one ico file. That way, the app shortcut can display both versions of the icon displayed as an option to pick.

craigwi:
I agree with @ZanderBrown.  It wouldn't be WinFile with a new icon...

Techokami:
@ZanderBrown Higher resolution versions of the old icon would be a good idea. But there should also be *lower* resolution icons provided as well. On a normal display without high DPI, you need a 16x16 version for the window title bar and, if people are using small icons on the taskbar (like me), the taskbar. Since there's nothing smaller than 32x32 in the original icon (it's only 32x32 in 16 color and 32x32 in black & white), Windows will try to compress the icon by itself. The end result looks like trash. That was the primary reason I did this, because this icon provides the extra sizes needed while not deviating too far from the original icon concept (a file cabinet). I think it might be because the color is different that it seems wrong?

@NazmusLabs Actually there is. See, icon files are indeed a container of multiple images, but it's supposed to be the same image at different sizes and color depths to adapt to the usage scenarios it would be put under. In Windows 3.0, this was simply a 32x32 pixel 16-color icon, and a 32x32 pixel 2-color (black & white) icon. In Windows 95, this was changed to a 32x32 pixel 16-color icon and a 16x16 pixel 16-color icon, because of the introduction of the Taskbar and the new usage of program icons in the window title bar. Windows 98 added the same sizes in 256-color format, and Windows XP added larger sizes and true color icons. But the icon file still represents one icon. If you want multiple icons in a single file, you would need to use an icon container. So to have both icons, you would need to add both as resources in the same program file.

@craigwi yeah I figured this is what would happen. Though if I were to try again with something more along the lines of the WinFile icon style, just in the extra sizes needed for modern Windows, would that be cool, or would that still be looked down upon?

ZanderBrown:
[SVG available here](https://en.wikipedia.org/wiki/File:WinFile_Icon.svg) so we can go infinitely 'big', smaller maybe more of a challenge

Techokami:
Actually I've been working now on some larger icons, I just finished a 48x48 version: 
48x48 WinFile

As well as a 16x16 version: 16x16 WinFile

I also got a 64x64 version, which I scaled down for the 48x48 version.

NazmusLabs:
@Techokami exactly. It's why I was a bit confused that you had multiple Icon files rather than one icon set.

@craigwi while I completely agree with you, we do need to supply larger versions of the icon, as otherwise File explorer and the desktop can't display the icon properly and has an ugly looking square transparent boarder around a tiny icon. This is especially important for my high DIP display where I prefer using the large icon view on my desktop.

Techokami:
Okay I've made a new PR from a clean fork, see #73

stuaxo:
This looks really nice.   The 16x16 seems to have a bit of a harsh shadow between the two drawers is the only thing I can see.

Amaroq-Clearwater:
The dithering on the icon is pretty harsh. I cleaned up the High-Resolution version in my own pull request.

PR #41, opened 2018-04-10
Code fixes

fuzun:
Some quick code fixes based on static analyser outputs.


fuzun:
I have splitted the commit and opened a different pull request #55

PR #43, opened 2018-04-10
C++ fixes: Reduce manual dynamic allocation, use std::algorithm

qtoncoding:
1. Reduce dynamic allocation of std::vector.
2. Change parameters of functions to use const& instead of pointer where applicable.
3. Use std::vector::push_back instead of insert at end.
4. Use std algorithms where applicable.


craigwi:
It looks like you forgot to include other changes.  I get a compile error:

Error	C2039	'back_inserter': is not a member of 'std'	Winfile	c:\users\craig\source\repos\winfile\src\wfgoto.cpp	105

qtoncoding:
You are correct, my bad. I forgot to add the change to BagOfValues.h as well. 
Fixed in a new commit.

craigwi:
Really nice changes.  However, In a simple perf measurement on my box the startup of WinFile from initializatin to the File.Goto cache working went from 1m 32s in the current build to 1m 52s with your changes.  Can you measure the difference on your system and report back?  Thanks.

qtoncoding:
I've done some preliminary testing on my end and find no noticeable difference. However, my C drive is an SSD and might have a simpler tree structure than your dev box.
One thing remains is that most of the time is spent in WFFindFirst, which is then spent in FindFirstFileEx. This is just the cost of doing this syscall, the rest of the code don't seem to take any time at all.
Here are some graphics I got from VS's profiler.

Master vs Pull Request on Debug Build, MSVC v140 (VS2015)
master-vs-pr-debug

Master vs Pull Request on Release Build, MSVC v140 (VS2015)
master-vs-pr-release

Master vs Pull Request on Debug Build, MSVC v141 (VS2017)
master-vs-pr-debug-vs2017


Master vs Pull Request on Release Build, MSVC v141 (VS2017)
master-vs-pr-release-vs2017

There isn't significant difference in Release builds, however, there's a 10 seconds gain in Debug build from VS2015 to VS2017. Kudos to MSVC team!

src/wfgoto.cpp
```diff
@@ -706,11 +685,11 @@ StartBuildingDirectoryTrie()
...
-		0L,
-		BuildDirectoryTreeBagOValues,
-		NULL,
-		0L,
-		&dwIgnore);
+							   0L,
+							   BuildDirectoryTreeBagOValues,
+							   NULL,
+							   0L,
+							   &dwIgnore);
```
ibvrubleg: IMHO, original formatting was much better here.

src/BagOValues.h
```diff
@@ -39,7 +39,7 @@ class BagOValues
 		wstring lowered;
 		lowered.resize(key.size());
 		transform(key.begin(), key.end(), lowered.begin(), ::tolower);
-		m_Values.insert(m_Values.end(),  make_pair(lowered, value));
+		m_Values.push_back(make_pair(lowered, value));
```
leeter: Why not use `emplace_back` instead? That would construct in place and prevent a copy?

src/wfgoto.cpp
```diff
@@ -95,61 +96,55 @@ bool CompareNodes(const PDNODE& a, const PDNODE& b)
...
 {
-	if (parents == NULL || children == NULL)
-		return NULL;
-
-	vector<PDNODE> *results = new vector <PDNODE>();
+	vector<PDNODE> results;
 
 	// for each child, if parent in parents, return
-	for (int i = 0; i < children->size(); i++)
+	std::copy_if(children.begin(),
```
leeter: why not use `std::begin` and `std::end` here so that if the type of `children` changes the code doesn't have to? Ditto below on 110.

src/wfgoto.cpp
```diff
@@ -95,61 +96,55 @@ bool CompareNodes(const PDNODE& a, const PDNODE& b)
...
+				 children.end(),
+				 std::back_inserter(results),
+				 [&parents](PDNODE const& child)
```
leeter: If we're in C++14 mode why not use `const auto&`? Does having the type here provide value?

src/wfgoto.cpp
```diff
@@ -95,61 +96,55 @@ bool CompareNodes(const PDNODE& a, const PDNODE& b)
...
 {
+	vector<PDNODE> result;
 	int count = trees.size();
 
 	if (count == 0)
-		return NULL;
+		return result;
+
+	// If any tree is empty, return empty
+	if (std::any_of(trees.begin(), trees.end(), [](auto& tree) { return tree.size() == 0; }))
```
leeter: Same as above concerning but `std::cbegin` and `std::cend` instead since this can be const. Also shouldn't it be `const auto&` and use `tree.empty()`?

src/wfgoto.cpp
```diff
@@ -95,61 +96,55 @@ bool CompareNodes(const PDNODE& a, const PDNODE& b)
...
-	vector<PDNODE> *outputA = new vector <PDNODE>(maxOutput);
-	vector<PDNODE> *outputB = NULL;
+	vector<PDNODE> outputA(maxOutput);
+	vector<PDNODE> outputB(maxOutput);
 	vector<PDNODE> *combined = NULL;
 	size_t lastOutput = 0;
 
 	// first is left side of merge; changes each time through the loop
-	vector<PDNODE> *first = NULL;
+	vector<PDNODE>* first = NULL;
```
leeter: shouldn't this be `nullptr`?

src/wfgoto.cpp
```diff
@@ -162,29 +157,27 @@ vector<PDNODE> *TreeIntersection(vector<vector<PDNODE> *> trees)
 		if (i == 1)
 		{
 			// on first time through loop, output is A and 'first' is trees[0];
-			first = trees[0];
+			first = &trees[0];
```
leeter: Consider using `.data()` instead if you can enable c++17 mode and `/permissive-`

src/wfgoto.cpp
```diff
@@ -409,18 +396,18 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes,
 	return TRUE;
 }
 
-vector<PDNODE> *GetDirectoryOptionsFromText(LPTSTR szText, BOOL *pbLimited)
+vector<PDNODE> GetDirectoryOptionsFromText(LPTSTR szText, BOOL *pbLimited)
 {
 	if (g_pBagOCDrive == NULL)
-		return NULL;
+		return vector<PDNODE>();
```
leeter: Any specific reason that `{}` won't work here?

src/wfgoto.cpp
```diff
@@ -482,17 +462,17 @@ VOID UpdateGotoList(HWND hDlg)
...
 	DWORD dw = GetDlgItemText(hDlg, IDD_GOTODIR, szText, COUNTOF(szText));
 
-	vector<PDNODE> *options = GetDirectoryOptionsFromText(szText, &bLimited);
+	vector<PDNODE> options = GetDirectoryOptionsFromText(szText, &bLimited);
 
 	HWND hwndLB = GetDlgItem(hDlg, IDD_GOTOLIST);
 	SendMessage(hwndLB, LB_RESETCONTENT, 0, 0);
 
-	if (options == NULL)
+	if (options.size() == 0)
```
leeter: Consider using `.empty()` instead

src/wfgoto.cpp
```diff
@@ -482,17 +462,17 @@ VOID UpdateGotoList(HWND hDlg)
...
 		return;
-		
-	for (int i = 0; i < 10 && i < options->size(); i++)
+
+	for (auto i = 0u; i < 10u && i < options.size(); i++)
```
leeter: Be extremely careful with unsigned iterators, these can cause undefined behavior.

src/wfgoto.cpp
```diff
@@ -482,17 +462,17 @@ VOID UpdateGotoList(HWND hDlg)
...
 	{
-		GetTreePath(options->at(i), szText);
+		GetTreePath(options[i], szText);
```
leeter: `operator[](int)` is not the same as `.at(int)` is this intended? This change removes bounds checking.

qtoncoding:
Lot's of fair review and comments, I'll address in a commit today :)

src/BagOValues.h
qtoncoding: I was under the impression that push_back can handle rvalue properly, but better safe than sorry I guess. Thanks for the feedback.

leeter: Honestly I'd be more worried about quirks in `std::pair` which honestly is probably not the right tool for this job. I'd actually suggest using a custom `struct` but that seems a bit invasive at the moment. You can also probably `std::move` `lowered` in as well as `value` if you're not using them elsewhere. That could remove an allocation and a copy for each.

qtoncoding: I agree with using `std::move` for `lowered` and `value`. A custom struct might be a bigger change than I intend to do at this time. I do intend to make bigger changes, but probably after talking with the author to get a clear understanding of his intentions.

craigwi:
@qtoncoding, thanks for doing the detailed perf analysis.  Can you resolve the conflict with wfgoto.cpp?  And revert the formatting changes in the file headers?  Thanks.

@leeter, any other comments?

src/wfgoto.cpp
qtoncoding: I think "address of first element" expresses the intent better than .data(), even though it's the same thing

src/wfgoto.cpp
qtoncoding: Can you elaborate on the UB in this case? I was under the impression that because we're comparing against size_t from .size(), using unsigned would please the compiler.

qtoncoding:
I have addressed the comments, as well as make a minor change (LPTSTR to LPCTSTR) to compile with the permissive- flag

craigwi:
We are really close.  Are the changes to WinFile.vcxproj necessary?  I want to leave the tool at v140.  Is the language standard 'stdcpplatest' necessary?

leeter:
@craigwi I'll be removing myself from any future contribution on this project. But you should be aware that's to enforce conformance so this can be cross compiler safe. It disables a lot of unnecessary and unused extensions and warns about using deprecated behavior. As for toolset v140 I would highly suggest moving to something newer as that has several serious and highly concerning compiler issues and does not support mitigating specter and meltdown and lacks support for control flow guard.

qtoncoding:
I can remove the changes to the project file. Despite that, I'll continue to work on future changes using v141 and at least the /permissive- flag, if not stdcpplatest. This is both for quality purpose, as well as supporting the conformance effort of the MSVC team.

qtoncoding:
Project file changes have been removed.

craigwi:
@DBJDBJ, let's discuss that in another issue.

@qtoncoding, thanks!

I am also open to discussing moving the project forward to VS2017 tools.  My question above was about whether the changes were *required* for this PR.  My preference is to keep such distinct changes separate.

PR #46, opened 2018-04-10
When cold starting on Win10 with multi monitors, starts basically invisible

tig:
I found on two different Win10 machines with multi-mon, the first time I ran winfile.exe the window was basically just the title bar (zero height client area) and infinitely wide (actually 0x80000000)).

This one line I removed is the cause and I can't see why it would be needed. SystemParametersInfo(SPI_GETWORKAREA) returns the correct RECT.

Did not test on any version of Windows other than Win10...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -286,3 +286,6 @@ __pycache__/
 *.btm.cs
 *.odx.cs
 *.xsd.cs
+
+# Visual Studio temp project files
+*.vcxproj
--- a/.gitignore
+++ b/.gitignore
@@ -286,6 +286,3 @@ __pycache__/
 *.btm.cs
 *.odx.cs
 *.xsd.cs
-
-# Visual Studio temp project files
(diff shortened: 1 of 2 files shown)


craigwi:
Hey Charlie, I thought I tested that, but glad it was a simple fix.

Can you revert the change to .gitignore?  I'll merge your request after that.

tig:
Sorry, my git skillz are lame. I couldn't figure out how to revert the .gitignore change (I actually thought I *had* reverted it before the pull request). I just copied the original over my edited version and added/commited that.

x4m3:
@tig to revert a commit, from the command line, you run `git revert xxx`, where `xxx` is the commit id, you can find it when you run `git log`.

hope it helps in the future :)

craigwi:
Right fix is more complicated.  See my change.  Thanks Charlie!

PR #48, opened 2018-04-10
revert Trie and add a comment

jsoref:
per @rbeesley in https://github.com/Microsoft/winfile/commit/7185e43f5458f581f8de6ee40780ddf4affb15f7#commitcomment-28501075


T8z5h3:
why was that changed?

PR #55, opened 2018-04-10
Code Improvements

fuzun:
Some quick code fixes based on static analyser outputs.

~~Resolves most of the warnings.~~ (Already addressed by d0493ea)


src/res.rc
```diff
@@ -7,9 +7,29 @@
...
+
+#ifdef WINVER
+#undef WINVER
+#endif
+#define WINVER 0x0400
+
+#ifdef _WIN32_WINNT
+#undef _WIN32_WINNT
+#endif
+#define _WIN32_WINNT 0x0403
```
leeter: Windows XP is not a supported platform by Microsoft I do not believe this is a valid change

fuzun: This was added to suppress the _WIN32_WINNT redefinition warning. The value was in the resource related section (resource predefinitions) in the project file.
They are most likely not needed because the recent change (commit: d0493ea) got rid of them and made it to use higher scoped corresponding values.

I will revert this commit along with other commits which were addressed by the latest changes.

craigwi:
I took several changes you proposed.  I don't see the need for the rest. Can you review and resubmit if you think they are needed?

Thanks!

craigwi:
Hi, per https://stackoverflow.com/questions/13147790/tilde-c-unsigned-vs-signed-integer it is true that ~0 is a signed integer, but since this code base assumes two's compliment representation, I don't see the issue.

The only change that makes sense to me is to change ~0 to ~0u.  But then you would have to go look at the other 5 similar cases.

craigwi:
Also, I have reverted the change to wfdir.c (in 0e36f9fc087ad5970a732e0bd87d89c8e68ea827) since that resulted in 'ld' for the size field instead of the size itself.

If you have a PR for the suggested changes above (~0 -> ~0u and related changes), open a new one.

PR #59, opened 2018-04-10
Restore XP compatibility

roytam1:
- Downgrade C/C++ syntax to VC2010
- Dynamic loading of Wow64 functions
- Create wrapper of RegGetValue

### DIFF
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -18,6 +18,35 @@
 
 BOOL IsFATName(LPTSTR pName);
 
+typedef BOOL (WINAPI *tW64P)(HANDLE, PBOOL);
+typedef BOOL (WINAPI *tFSDisable)(PVOID*);
+typedef BOOL (WINAPI *tFSRevert)(PVOID);
+
+HMODULE hKernel = NULL;
+BOOL bIsWow64 = FALSE;
+BOOL bWow64FuncsSearched = FALSE;
+
+tFSDisable pDisableFunc = NULL;
+tFSRevert pRevertFunc = NULL;
+tW64P pIsWow64Func = NULL;
+
+void _InitWow64Funcs() {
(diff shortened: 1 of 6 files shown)


leeter:
I don't think this should be merged into master, it significantly holds back the rest of the codebase from using features in C++ that would increase safety and produce a faster better result.

NazmusLabs:
We should combine efforts and use this on a separate retro branch, as others noted.

See issue #3

That way that specific branch can be dedicated to work on as many legacy versions of Windows as possible.

Perhaps we can call it Version 9.x

Azarien:
You don't need to downgrade to VS2010, because you can still use VS2017 to compile XP applications.

roytam1:
@Azarien because I want to dogfood in XP and VC2010 is last version working on XP.

NazmusLabs:
@Azarien Using older versions of VS will not only help @roytam1 help with XP dogfooding but using older libraries makes it easier for it to support even older versions of Windows, such as 2000, and 9x, which is the aim for a retro branch.

NazmusLabs:
Should this PR be merged into the retro branch? Currently the solution in the retro branch is still using VS2015. Or do we want to use versions of visual studios that are older than 2010, such as VS2008 or VS2005.

NazmusLabs:
@roytam1 yes, exactly. The idea is to those changes into the retro branch. That's why the branch was created.

craigwi:
Has anyone tried the Win XP support in VS 2017?  Perhaps that's a way to get an XP build?

craigwi:
Still wondering...

roytam1:
@craigwi this PR may need to be closed in favor of retro branch instead?

craigwi:
OK; thanks for responding.

PR #61, opened 2018-04-10
Add Bash Shell support (partially fixes #44)

### LINKED ISSUE #44: Make "Start Bash Shell" feature configurable
The File Manager supports "Start Cmd Shell" and "Start PowerShell". Can we add "Start Bash Shell", if users have installed Bash on Windows?

Edit:
* [X] Add "Start Bash Shell" (Done in #61)
* [ ] Make this feature configurable

xiaoyinl:
This PR adds a "Start Bash Shell" to menu, which opens Bash on Windows, if it's installed.

While I totally agree with the comments in #44 that it is better to make this configurable, I don't have enough Windows GUI knowledge to write a configuration dialog. So far I only implemented a menu item "Start Bash Shell", which launches `bash.exe` if it exists. I want to share this code so that, if you want, you can develop on top of this PR.

### DIFF
--- a/src/res.rc
+++ b/src/res.rc
@@ -58,6 +58,7 @@ BEGIN
 	"^x",       IDM_CUTTOCLIPBOARD, NOINVERT
 	"^c",       IDM_COPYTOCLIPBOARD, NOINVERT
 	"^v",       IDM_PASTE, NOINVERT
+	"^b",       IDM_STARTBASHSHELL, NOINVERT
 	"^k",       IDM_STARTCMDSHELL, NOINVERT
 	"^p",       IDM_STARTPOWERSHELL, NOINVERT
 	"^g",       IDM_GOTODIR, NOINVERT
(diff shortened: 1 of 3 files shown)


NazmusLabs:
I'd love this to be included in the interm until a configuration system is developed. Having three most used shells isn't a bad idea.

I'm, curious though, if this throws an exception in older versions of Windows that predates WSL

xiaoyinl:
@NazmusLabs It won't throw an exception if WSL doesn't exist. It just won't do anything if it can't find `bash.exe`

NazmusLabs:
@xiaoyinl Yeah, I realized that after I checked the code. :)

I did see that you did hard code the system32 folder so that the 32 bit version also works. I'm curious how this affects OSes prior to XP 64 bit, which I don't believe knows what sysdrive is, but I could be wrong.

xiaoyinl:
@NazmusLabs I am not sure if it understands it or not. But that doesn't matter in this case, because if it doesn't understand, PathFileExists will return false, which is correct anyway, because Bash only runs on Windows 10 x64.

Issue #53, opened 2018-04-10
In executable properties the product name is "Win7 DDK driver"


Anixx:
In executable properties the product name is "Windows (R) Win7 DDK driver"

NazmusLabs:
Wow, interesting...I wonder what the device driver kit for Windows 7 have to do with Winfile...

ZanderBrown:
Probably just copy pasted the manifest

matthewjustice:
This should be fixed now in master with PR #89

genghaibo:
8  src/Winfile.vcxproj
 @@ -109,7 +109,7 @@
        <AdditionalIncludeDirectories>.;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
      </ResourceCompile>
      <Link>
 -      <AdditionalDependencies>ole32.lib;shell32.lib;shlwapi.lib;comctl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
 +      <AdditionalDependencies>ole32.lib;shell32.lib;shlwapi.lib;comctl32.lib;mincore.lib;%(AdditionalDependencies)</AdditionalDependencies>
        <AdditionalLibraryDirectories>
        </AdditionalLibraryDirectories>
        <GenerateDebugInformation>true</GenerateDebugInformation>
 @@ -134,7 +134,7 @@
        <AdditionalIncludeDirectories>.;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
      </ResourceCompile>
      <Link>
 -      <AdditionalDependencies>ole32.lib;shell32.lib;shlwapi.lib;comctl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
 +      <AdditionalDependencies>ole32.lib;shell32.lib;shlwapi.lib;comctl32.lib;mincore.lib;%(AdditionalDependencies)</AdditionalDependencies>
        <AdditionalLibraryDirectories>
        </AdditionalLibraryDirectories>
        <GenerateDebugInformation>true</GenerateDebugInformation>
 @@ -167,7 +167,7 @@
        <AdditionalIncludeDirectories>.;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
      </ResourceCompile>
      <Link>
 -      <AdditionalDependencies>ole32.lib;shell32.lib;shlwapi.lib;comctl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
[...]

PR #67, opened 2018-04-10
Specify product name rather than rely on ntverp default

matthewjustice:
Issue #53 calls out that the Winfile.exe product name is "Windows (R) Win7 DDK driver". This is because res.rc includes ntverp.h, and the version of ntverp.ver that is included with Windows 10 SDK defines the following:

`#define VER_PRODUCTNAME_STR         "Windows (R) Win 7 DDK driver"`

I addressed this by defining the product name as "File Manager", although another product name would be fine if something else is preferred. I also undefined VER_PRODUCTNAME_STR first to avoid resource compiler warning RC4005.

### DIFF
--- a/src/res.rc
+++ b/src/res.rc
@@ -688,6 +688,7 @@ END
 #define VER_FILEDESCRIPTION_STR     "File Manager"
 #define VER_INTERNALNAME_STR        "fileman\0"
 #define VER_ORIGINALFILENAME_STR     "WINFILE.EXE"
+#undef  VER_PRODUCTNAME_STR
 #define VER_PRODUCTNAME_STR         "File Manager"
 
 #include "common.ver"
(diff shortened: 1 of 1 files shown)


craigwi:
I would like to combine the AboutDlg changes with this change and do proper build versions.

I am also looking at AppVeyor for build/deploy via Chocolatey.

Anyone have experience with https://github.com/GitTools/GitVersion or recommend something else?

matthewjustice:
I'm working on a change right now that removes the ntverp.h dependency (since it is primarily intended for internal NT builds), and defines the full version, etc in the resource file, and then displays the version info in the AboutDlg. As it stands right, the version number is fully coming from ntverp.h, which we don't want. 

So, I could go ahead and make this change as first step (break the ntverp.h dependency, define version info in the rc file, and display the version in About), and then we could look into ways to automate updating the version number as part of the build process.

matthewjustice:
PR #89, if accepted, will supersede this PR.

craigwi:
I think PR #89 a great change.  I will merge it after I undo the last commit.

PR #42, opened 2018-04-10
build on mingw compiler

mattn:
include some fixes

PR #68, opened 2018-04-10
Port to mingw32 and add a Makefile

ebraminio:
This just works but the patch quality should be improved a lot before considering for merge obviously but I like to know if the upstream will see any future for such a thing :) Thanks!


ZanderBrown:
#42

ebraminio:
Oops! Thanks!

PR #70, opened 2018-04-10
Fix most compilation warnings at /W3 level

Ybalrid:
This patch fixes a number of simple warnings at /W3 level

They mostly boils down to theses simple (and not too severe) errors:

 - Storing unsinged numbers in singed variables, then comparing them
 back to unsigned numbers
 - Declaring for loop index variables as signed when testing it against
 unsigned values (eg: size of a container)
 - Using undeclared funcions in C. Compiler will fallback to thinking of
 them as extern function returning `int`. This *happens* to work because
 the linker will find the correct function later.

 **Note**: This path doesn't replace the usage of "unsafe" C standard function
 with their "safe" (generally postfixed _s) counterpart. I'm planning to
 submit a separate PR for that ;-)

Signed-off-by: Arthur Brainville (Ybalrid) <ybalrid@ybalrid.info>

### DIFF
--- a/src/treectl.c
+++ b/src/treectl.c
@@ -2194,7 +2194,7 @@ TreeControlWndProc(
       WCHAR    ch;
       PDNODE    pNode;
       WCHAR rgchMatch[MAXPATHLEN];
-      INT cchMatch;
+      SIZE_T cchMatch;
 
       //
       // backslash means the root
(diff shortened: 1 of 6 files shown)


craigwi:
Thanks.

Issue #52, opened 2018-04-10
Default icon in the menu bar


Anixx:
In the beginning of the menu bar there is a default icon with seemingly no purpose.
![clipboard_20180410](https://user-images.githubusercontent.com/1271736/38552437-a25f4b4a-3cab-11e8-931e-5a186c3a5f68.png)

ZanderBrown:
WinFile is an MDI application

Apparently you have a Maximised subwindow (you should have double window controls on the left as well)

Back in the day windows didn't have icons so all was fine:
![screenshot_boxes-unknown-10-clone_2018-04-10_12 28 21](https://user-images.githubusercontent.com/17074021/38552799-3d3ef634-3cb5-11e8-8e85-423398478649.png)
However they did have icons for there minimised state:
![screenshot_boxes-unknown-10-clone_2018-04-10_12 29 11](https://user-images.githubusercontent.com/17074021/38552803-41d874f4-3cb5-11e8-834f-ae6829d198cd.png)

Which still works in this release (running under wine):
![screenshot from 2018-04-10 12-31-58](https://user-images.githubusercontent.com/17074021/38552934-9c22e1b0-3cb5-11e8-8172-114be0a36a66.png)
(ico is [here](https://github.com/Microsoft/winfile/blob/master/src/wftrdir.ico))

So I assume the API for setting the Window icon was change around Win95 and WinFile didn't keep up, should be a quick fix for someone

PR #72, opened 2018-04-10
Set the small icon for the MDI child on create window. This should di…

neilgierman:
…splay the TreeDir icon instead of the default windows icon.

Should resolve https://github.com/Microsoft/winfile/issues/52


craigwi:
There are three icons for the child windows.  This fix properly sets the icon depending on whether the child window has one or both sub-windows.

PR #73, opened 2018-04-10
Updated icon, take 2

Techokami:
This is the WinFile icon, updated for the larger array of icon sizes used by modern Windows. Still contains the original 32x32 icon artwork! I just added a 16x16, 48x48, and 64x64 version to make things look nicer in modern environments and high DPI displays. Also contains appropriate black & white (2 color) versions to go with the 16 color versions.


ZanderBrown:
I'm not sure because by the looks of it you have repeated the pattern to fill in the gaps for the larger sizes, while this is fine I'd have gone for the 1px=4px technique where we effectively make the pixels bigger for larger resolutions. Thus giving us a large scale sharp pixilated icon.

Great Artwork though

NazmusLabs:
@ZanderBrown I'm fine either way, but I do think this looks a lot cleaner and goes well with all the other app icons I have on my desktop.

Techokami:
I went with the same art style used in Windows 95 to scale up icons used for the side banner of Wizards. So, it still has the same style and limitations, just bigger. Just making the icon 2x resized is, imho, a lazy cop-out... but then again so was my first PR attempt 😏

Thank you for accepting my work!

craigwi:
Nice!

Techokami:
Thank you so much, Craig! 😄 
I'll keep tabs on the work being done for high DPI support. I can do the same for the rest of the needed artwork if you'd like!

NazmusLabs:
@Techokami, 64x64 is not sufficient for Windows 10, and High DPI displays. Windows 10 takes advantage of 128x128 and 256x256. It allows Windows 10 icons to look EXTREMELY SHARP and clear.. Here's how it looks if icon size is set to large. Note, it isn't even set to extra large, which would cause it to look even worse. But the work you did DOES make the icon fit nicely if you leave the icon sizes to medium, which is definitely nice and appreciated!

high res icons in windows 10


I want to take a moment to say that I appreciate the icon work that's been done in Windows 10. It's not something we notice everyday, but if you do see the actual icons that are used in Windows 7 versus that in Windows 10, the difference is night and day! The following screenshot shows how crystal clear the Windows 10 icons really are compared that of Windows 7:

windows 7 vs 10 icons

Techokami:
@NazmusLabs  Oh yikes, I should go make some higher res versions!

ZanderBrown:
Of course our icon is pixilated by design making this an interesting discussion

Techokami:
I've submitted a PR (and it got merged!) for higher resolution icons. NOW it goes to 256x256 😃

NazmusLabs:
Nice!

Umm, this might be a newbe question for me, but I appreciate the help: I was trying to figure out how to compile this in VS, as I could see how project or solution file. I am guessing I'd need to use the "open folder" option in VS, or am I supposed to do something else. Thanks again!

ZanderBrown:
The solution is in src, probably should be in root but for now that's where it is

Issue #40, opened 2018-04-10
Create WinFile specific About box


random-ideas:
Perhaps I'm off the mark on the intent, but the Help > About box shows winver but without user or org information (left in this). Super cool otherwise though!
![image](https://user-images.githubusercontent.com/37677613/38530126-2e9febfa-3cac-11e8-88f4-e93f6ebe0352.png)

craigwi:
Read the source; wfcommand.c line 2083 is unchanged from the past and calls -- wait for it -- ShellAbout().

I do agree, however, that WinFile should have a proper about box.  Any takers?

ZanderBrown:
To be in keeping with the original it would be great if someone could recreate the Win3X era ShellAbout
![screenshot from 2018-04-10 12-04-39](https://user-images.githubusercontent.com/17074021/38551695-c599e7c2-3cb1-11e8-9148-77089177226b.png)

daniel-white:
I noticed they were different. Perhaps it has something to do with a manifest.

Anixx:
On Win 8.1 it reports incorrect Windows version (6.2 instead of 6.3) and wrong build number.

ZanderBrown:
@random-ideas @daniel-white @Anixx are you running in compatibility mode?

That would explain wrong build ext

Anixx:
No. No compatibility mode. In Win8.1 the APIS had changed, all old progs show windows version as 6.2 build 9200 while newer programs show 6.3 build 9600

craigwi:
I renamed the issue.  Feel free to submit a PR.

matthewjustice:
I have a change almost ready for a simple about box, but I'm not sure what text to include. Below is a screenshot of what I have now. Suggestions before I submit a PR? Or I could just submit as-is and let others update the text as needed.

<img width="283" alt="winfile-about" src="https://user-images.githubusercontent.com/18629374/38585105-a6d4106c-3ce6-11e8-846d-dfc980544aea.png">

ZanderBrown:
It's a good starting point

matthewjustice:
@ZanderBrown - Thanks. PR #76 submitted.

craigwi:
@matthewjustice's PR addresses this for now.

PR #76, opened 2018-04-11
Add WinFile-specific about dialog box

matthewjustice:
As recommended in issue #40, an "About" dialog was added that is specific to the application. The dialog is simple; it could be expanded upon as needed. 

One note of interest: winfile.h already contained a function prototype named "AboutDlgProc", seemingly a dialog proc callback for an "About" dialog, but the code did not contain an implementation of that function. I suspect that the code previously had such a function and it was removed, or it was never implemented. Regardless, I left that prototype in place and implemented the DialogProc callback function using the same function name and signature.

### DIFF
--- a/src/wfcomman.c
+++ b/src/wfcomman.c
@@ -2079,8 +2079,7 @@ ACPCallHelp:
        break;
 
     case IDM_ABOUT:
-       LoadString(hAppInstance, IDS_WINFILE, szTitle, COUNTOF(szTitle));
-       ShellAbout(hwndFrame, szTitle, NULL, NULL);
+       DialogBox(hAppInstance, (LPTSTR)MAKEINTRESOURCE(ABOUTDLG), hwndFrame, (DLGPROC)AboutDlgProc);
        break;
 
     case IDM_DRIVESMORE:
--- a/src/wfdlgs.h
+++ b/src/wfdlgs.h
@@ -61,6 +61,7 @@
 #define COMPRESSERRDLG          59  //  Compression Error Dialog
 
 #define GOTODIRDLG    60
+#define ABOUTDLG      61
 
 #define IDD_TEXT      -1
 #define IDD_TEXT1     100
--- a/src/wfdlgs3.c
+++ b/src/wfdlgs3.c
@@ -724,6 +724,28 @@ DoHelp:
   return TRUE;
 }
 
+/*--------------------------------------------------------------------------*/
+/*                                                                          */
+/*  AboutDlgProc() -  DialogProc callback function for ABOUTDLG             */
+/*                                                                          */
+/*--------------------------------------------------------------------------*/
+
+INT_PTR
+AboutDlgProc(register HWND hDlg, UINT wMsg, WPARAM wParam, LPARAM lParam)
+{
+    switch (wMsg)
+    {
+    case WM_COMMAND:
+        switch (GET_WM_COMMAND_ID(wParam, lParam))
+        {
+        case IDOK:
+            EndDialog(hDlg, IDOK);
+            return TRUE;
+        }
+    }
+    return FALSE;
+}
+
 VOID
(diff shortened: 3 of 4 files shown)


matthewjustice:
@NazmusLabs - That is good feedback; thanks! Let me speak to the specific points you raised:

- Version number, agreed that the about box should display it. However, I didn't want to hard-code it as a string. I'd rather see a major, minor, revision, build number in the binary, which would show up in the file properties as well as the about box. I didn't see such a versioning pattern already in place (maybe I missed it), so adding versioning seemed like a bigger effort for later.
- "(C) Microsoft Corporation, All rights reserved." - That does seem like standard text, but I was hesitant to make copyright claims without some input from Microsoft. @craigwi - Do you have any feedback on this point?
- Calling out the MIT License makes sense; good idea. @NazmusLabs were you recommending a link to the MIT license itself here or a link to the winfile project?

NazmusLabs:
@matthewjustice I got the copyright information from the License file both on the github and also included with the release binary :)

And for versioning, there isn't any version information in code because it relies on the NT version, which is why just displays the Winver dialog box. I believe Windows 3.1's version of WinFile did have its own about box. Was probably removed in 9x or NT.

I suggested 10.0. Rather than 10.0.0 for this very reason. The numbers after the minor version are typically build and revision numbers. And if I were going to hard code a version number, I'd just put Version 10, or version 10.0. It'll be a while before we'd consider calling it Version 10.1 or 11. So I thought hard coding version 10.0 for now was a safe bet.. By the time we have enough features to warrant a v10.1 release, I assumed we'd have a proper version string and the about box would have been updated.

matthewjustice:
I've made two changes to this PR:

1. The text is updated as discussed. See screenshot below.
2. I'm now handing IDCANCEL in the dialog proc in case the user clicks the X in the upper right. That should have been there before.


about2

Some things to note: 

- I did go with the hard coded "10.0" string for now, but using the [VS_FIXEDFILEINFO](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646997(v=vs.85).aspx) approach is something I (or someone else) should consider for later, along with general versioning strategy.
- The link to GitHub isn't a hyperlink. Adding a [SysLink](https://msdn.microsoft.com/en-us/library/windows/desktop/bb760706(v=vs.85).aspx) requires some other changes and I wanted to keep things simple.

NazmusLabs:
The dialog looks great! And what would happen if the user clicked on the x before, haha.

And, right, I forgot to ask about the link. I was going to ask that I didn't believe that COM APIs WinFile is using had support for hyperlinks within in dialog boxes. I remember that becoming a thing with dot net apps and then later with Vista, where native Windows dialog boxes has better support for hyperlinks.

I think it's be easier, for now, to have a button beside the "OK" button labeled "License" that loads the local license text file that's includes with the app.

matthewjustice:
Thanks @NazmusLabs! If the user clicked the X before, nothing would happen, so not a great user experience! Yes, dialog hyperlinks via SysLink haven't always been there; the SysLink control is in ComCtl32.dll v6, and requires a manifest declaration. I like your license button idea... maybe something for a future PR?

craigwi:
nice work; many thanks!  Made a few tweaks and added the about box.

matthewjustice:
@craigwi  - I'm glad to help, and I'm glad you could use it! One small thing I noticed about one of the tweaks, the WM_INITDIALOG code path will result in AboutDlgProc returning FALSE, which in theory will "prevent the system from setting the default keyboard focus" ([MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms645428(v=vs.85).aspx)). Now, in practice, it doesn't seem to matter, and the code works fine.

craigwi:
Thanks; I'll fix that.

matthewjustice:
@craigwi - FYI PR #89 made changes in the WM_INITDIALOG code, so I added a fix for the return issue mentioned above.

PR #80, opened 2018-04-11
HDPI Icon

Techokami:
Added 96x96, 128x128, and 256x256 icons. Perfect for use with 4K displays.


Techokami:
Thinking about it, the high pixel density combined with the 16-color Windows icon art style from Windows 3.1/95 actually revives what some of the art tricks employed were supposed to do. The dithering of yellow and light gray is supposed to blur on CRT displays to create a darker yellow shade that isn't the dark yellow color in the palette. With crazy high pixel densities in modern smartphones, tablets, and 4K displays, the pixels are clumped so close together that they create the same effect.

PR #81, opened 2018-04-11
enable goto with x64

jamesmoore:
remove preprocessor directives and switch SetWindowLong for SetWindowLongPtr to enable x64 compatibility


craigwi:
@jamesmoore, thanks for tracking down that detail!

PR #82, opened 2018-04-11
Remove dependency of VC++ Runtime DLL

### LINKED ISSUE #24: A couple of suggestions and a small nit to pick
Your effort is to be praised, 

I've been using WinFile since it first came out, and still use it every day. Unfortunately it got broke in Vista and this guy figured out a fix for it: http://people.ee.ethz.ch/~davidsch/vistafm/

It was a bit fiddly to get it to work, but it did work under Vista, Win-7 and Win-10. But there were a couple of problems mainly dealing with permissions, and Winfile.ini not being saved every time. Specifically in the Windows32 directory, the ECT directory didn't show up in David's version. It shows up fine in yours. Another one was the WinEvt sub directory with the Event Logs. It didn't show up in Davids but shows up fine in yours. I'm sure there are others but it looks like your version solved all of the missing sub-directories.

The link given in the How to Geek article works: Visual C++ Redistributable for Visual Studio 2015 and you have the option of choosing a 32-Bit version or 64-Bit version. I chose the 64 bit version and got: **The program can’t start because MSVCP140.dll is missing from your computer**. 

But when I installed the 32-bit version, the error message went away and File Manager started right up. I'm running Win-7 and the error message might not appear if you are using Win-10. But it might be a good idea to let How to Geek know that they should mention to download the 32 bit version. 

Your version of File Manager worked flawlessly and as I used it, something was a bit different from the original version. I couldn't put my finger on it, and then I realized that it was how your version handled the display of the Directories and Sub-Directories. The original has a solid outline, and your version doesn't. It might have something to do with the newer libraries, dunno.

All in all, it's a welcome addition as far as I'm concerned.

Good job, well done.

file-manager-hidden-files

file-manager-folders

Biswa96:
This will remove the dependency of MSVCP140.dll and MSVCR140.dll. Hence users don't need to install  VS 2015 C++ runtime. This may fix #24.

### DIFF
--- a/src/Winfile.vcxproj
+++ b/src/Winfile.vcxproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|Win32">
@@ -97,7 +97,7 @@
       <PreprocessorDefinitions>DBG;_X86_;_M_IX86;WIN32;_WIN32;_WINDOWS;LFN;STRICT;FASTMOVE;LFNCLIPBOARD;W3;UNICODE;_UNICODE;USELASTDOT;MEMDOUBLE;WINVER=0x0601;_WIN32_WINNT=0x601;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <MinimalRebuild>true</MinimalRebuild>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
-      <RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>
+      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
       <PrecompiledHeader>
       </PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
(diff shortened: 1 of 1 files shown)


craigwi:
Seems simple enough.  Any downsides?

leeter:
@craigwi security would be a big one, along with conflicts with other CRTs that could be an issue

Biswa96:
I tested it in my Windows 10. But you may test in XP or Win7. If it is not compatible you may close PR.

leeter:
The more appropriate solution to this is an installer with bootstrapper that installs the appropriate CRT. That way we don't potentially leave users open to security vulns that can't be patched because they have an old version of the embedded CRT.

leeter:
@Biswa96 I don't think actually fixes the problem, because the embedded CRT may not support their OS. An installer would inform them at install time and allows for greater security overall.

Biswa96:
Ok, wait. I'll test that in XP, Win7, Win8.1 VM.

leeter:
That still does not fix the security concern.

Biswa96:
Should I close this PR?

craigwi:
not just yet.

craigwi:
It occurred to me that the suggestion by  @xiaoyinl to address @leeter's comments was to have an installer update things.  This mirrors @leeter's suggestion to ... have an installer the provide correct VC runtime.  

Furthermore, this is a one time problem.  Closing.

Biswa96:
Why did you close this PR? Is there any adverse effect with this change?  With this change you've not to include those DLLs and winfile becomes a standalone exe. The benefit is that no installer need.

NazmusLabs:
I agree. I think the installer should be part of the exe. But can you verify if the license for the redistribal allows you to just include the DLL as part of an app or does it requires you to distribute the installer. Since VC++ redistributal isn't free software, we need to be careful about how to distribute it.

craigwi:
I closed because the two primary distribution mechanisms have an "installer":

1. choco install winfile (works now)
2. Windows Store (future)

Both of those will install the VC runtime and then Microsoft gets to keep it up to date.

Random copying of winfile.exe does need to be accompanied by a onetime manual install of the runtime on relevant systems.  Perhaps the mingw version can be built differently.

ZanderBrown:
mingw version will likely be statically linked to its runtime so avoids vc++

PR #89, opened 2018-04-12
Add version information to RC file and display in About dialog box

matthewjustice:
This change does the following:

1. Removes the dependency on ntverp.h. Without this change, versioning is set by ntverp.h. This makes sense for building Windows where all binaries are versioned together, but that no longer applies to this code.
2. Versioning and related strings are now defined in res.rc. These values could be updated programmatically at build time if needed. This addresses issue #53 and supersedes the changes in PR #67.
3. Building on pull #76 for issue #40, The About Dialog now displays the VER_PRODUCTVERSION value that is contained in the VS_VERSION_INFO in the winfile.exe binary. This means that the values shown in explorer file properties will match what is in the About box. If we choose to update these values in some way other than the RC file, the about box will still pick up the correct values at runtime.

Notes:

- mincore.lib was added as a new linker dependency for GetFileVersionInfo  and related functions.
- strsafe.h was included to allow for StringCchPrintf
- Example of updating the product version below. The version below is just an example; in the code the version is 10.0.0.0 for now.
winfile-version


src/wfdlgs3.c
```diff
@@ -1699,3 +1714,63 @@ DestroyCancelWindow()
...
+        if (cbVerInfo)
+        {
+            pFileVerInfo = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, cbVerInfo);
+
+            if (pFileVerInfo)
+            {
+                if (GetFileVersionInfo(wszCurrentModulePath, 0, cbVerInfo, pFileVerInfo))
+                {
+                    // Get the pointer to the VS_FIXEDFILEINFO structure
+                    if (VerQueryValue(pFileVerInfo, L"\\", (LPVOID *)&pFixedFileInfo, &uLen))
```
xiaoyinl: Since you are using VerQueryValue (not VerQueryValueW), is there a reason why you use `L"\\"`, rather than `TEXT("\\")`?

matthewjustice: Good point. I was assuming UNICODE would be set at build time, but I agree that using TEXT would be more consistent. I'll make that change and also use TCHAR rather than WCHAR.

craigwi:
Hi @matthewjustice , can you update this PR now that I have reverted b51bf087dc86acab70b0ab5f439835e10efa6c2f?  Thanks.

matthewjustice:
@craigwi - I'm not sure if that is what you had in mind, but I was able to also revert b51bf08 in my about-version branch (see commit 71bce76). I think the changes in my branch should merge cleanly. I'll admit to not being an expert in Git, so let me know if I need to take a different approach.

craigwi:
Looks fine.  A squash and merge gives the right result.

craigwi:
Your change, squashed, is checked in.  I hope I didn't screw up again.

matthewjustice:
Thanks @craigwi! The commit looks good to me.

NazmusLabs:
So are we going to use gitversion to have the version number automatically updated with commits? It sounds the most practical.

PR #93, opened 2018-04-14
Add makefile for mingw

mattn:
This is branch of PR from #42

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ artifacts/
 *.ilk
 *.meta
 *.obj
+*.o
 *.pch
 *.pdb
 *.pgc
--- a/src/GNUmakefile
+++ b/src/GNUmakefile
@@ -1,0 +1,62 @@
+SRCS = \
+	dbg.c \
+	lfn.c \
+	lfnmisc.c \
+	numfmt.c \
+	suggest.c \
+	tbar.c \
+	treectl.c \
+	wfassoc.c \
+	wfchgnot.c \
+	wfcomman.c \
+	wfcopy.c \
+	wfdir.c \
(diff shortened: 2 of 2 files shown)


mattn:
@craigwi If you want single commit, please let me known. I'll squash commits before you will merge.

craigwi:
Thanks @mattn.  There was a comment about the name of the makefile.  I would prefer to have the mingw suffix, but are others ok with that?

ZanderBrown:
I'm not sure I see the point of the prefix

tuldok89:
How about renaming Makefile.mingw to GNUmakefile?

ZanderBrown:
Doesn't make much difference, still have to manually specify the Makefile to `make`

tuldok89:
It does make a difference. GNU Make also reads GNUmakefile by [default](https://www.gnu.org/software/make/manual/make.html#Makefile-Names).

ZanderBrown:
Interesting but
> The first name checked, GNUmakefile, is not recommended for most makefiles. You should use this name if you have a makefile that is specific to GNU make

Is worth noting

tuldok89:
My point exactly. NMAKE doesn't understand this makefile, so I think it's fine to name it `GNUmakefile`.

ePirat:
The makefile does not look GNU make specific to me, so I dont think it should be named `GNUmake`.

mattn:
I renamed Makefile.mingw to GNUmakefile since this makefile use subst as ePirat said.

mattn:
I squashed commits to 1 commit. Please push merge button if you are okay. Thanks.

mattn:
Thanks all!

PR #95, opened 2018-04-14
Added download link to Readme

NazmusLabs:
I've seen people asking for a download link to the precompiled binaries. There's no obvious way to get to the releases page. Therefore, I download section in the main readme file so that users, especially those who are not developers and are not experienced with using GitHub can easily find the link to download the latest version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,32 @@
 # ![icon](winfile.png) Windows File Manager (WinFile)
 
 The Windows File Manager lives again and runs as a native x86 and x64 desktop app
-on all currently supported version of Windows, including Windows 10.
+on all currently supported version of Windows, including Windows 10. I welcome your thoughts, comments and suggestions.
 
-I welcome your thoughts, comments and suggestions.  There are two
-primary versions of the source code in the master branch:
+There are two primary versions of the source code in the master branch:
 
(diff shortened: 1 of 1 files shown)


ZanderBrown:
I thinks it's better just to link to the releases page

Biswa96:
In my opinion, it is nice to provide download details in readme file. Because users/newcomers first see the readme file, not the release page. So they get a summary/usage details (including download) of the project from that readme file.

NazmusLabs:
I did add a link to the releases page. There are a total of three links. :)

ZanderBrown:
@NazmusLabs I meant *only* that link so we don't have to keep manually updating the other 2

NazmusLabs:
@ZanderBrown I had initially thought about doing just that changed my mind. Besides, we would only need to update one link. The Original_Plus doesn't change. Only the latest stable. If we only lined to the releases page, the Original_Plus would get buried with each new version, making it confusing for people who wanted to get the vanilla binary.

As for having to update the link to the latest build, at this moment, it doesn't seem too tedious. We haven't even had a single release post launch. And I don't see any hints that we'll be cranking out a bunch of releases in a short period of time anytime in the near future. Finally, it's possible that after a new release, we may update other portions of the Readme anyway to reflect some of the notable new features.

These were the thought process that lead me to make it the way I did.

Issue #77, opened 2018-04-11
Add additional languages beyond English, Chinese, Hebrew, Japanese, Polish, German


Anixx:
Windows releases contained Winfile in multiple languages (such as Russian). This project currently has only English language version.

craigwi:
I'll go digging in the NT4 source tree for the other languages.  I'll need help with any new/changed strings.

magol:
I can help with the Swedish language version.

daviunic:
Considering Windows is now available in languages in which NT 4.0 wasn't, will there be any additional translation efforts for such languages?

ZanderBrown:
@daviunic you offering?

IQBigBang:
I could do Czech and Slovakia version if developer would want to translate...

thecatkitty:
I can help with the Polish resources.

I think it would be rather all languages in one EXE than additional MUI files, wouldn't it?

daviunic:
@ZanderBrown yes, I could do the Slovenian translation.

craigwi:
I have yet to find any translations of the WinFile strings inside Microsoft.

What do folks think about @thecatkitty's idea that all languages are in one EXE?

ZanderBrown:
Makes sense, perhaps we need additional UI to override system language?

Anixx:
I do not think this utility should override system language.

mischastik:
I'd be glad to help if assistance with German is needed.

ZanderBrown:
@Anixx to clarify:

WinFile should use the system language by default, as should all applications, but have the option to manually specify a language use

Anixx:
@ZanderBrown I know no app that provides this feature.

NazmusLabs:
@ZanderBrown I've seen plenty of apps that give you the option to switch to another language or use the default system language.

I'm not arguing either way in this case. I'm just noting that I've seen may apps that do this and many that don't.

This feature will definitely be useful for Windows 7 or older. Windows 7 doesn't let you switch languages on the fly unless you have ultimate. With windows 10, there are more languages to download on demand that ever, and the process is very easy now that it is distributed through the Windows Store, as of Version 1803.

Again, I'm neutral in regards to this point, as it doesn't affect me significantly and, thus, won't be fair for me to just pick a side for the same of it.

bitsbr:
I could try to help in translations to Portuguese and Brazilian Portuguese .

ttsirkia:
How the translation was originally done?

[...]

PR #101, opened 2018-04-15
Making Winfile translatable

peterooch:
Moved all code that can be translatable to winfile.dlg that got renamed to en-US.rc, and now res.rc includes the aforementioned file.
This enables translations that are mentioned at issue #77

Note: for some reason the japanese characters got garbled while making the commit...


peterooch:
@NazmusLabs Please try with the additional commit that i just added,
Its includes 2 strings from c57f75539c91c461754910850e3d8c7b01e65640

craigwi:
We have a couple different efforts on translation.  I definitely support a translated version, but it isn't clear the right direction, technically.  Can we get some more ideas?

mattn:
https://github.com/Microsoft/winfile/pull/120 has more translations, I guess.

peterooch:
This PR is "ground work" for translations, #120 has some similar changes, but differences in resource file structure/layout and also debate on how to store the files,
i have only done and prefer to edit resource files manually, not in the resource editor of VS,
in there is a different approach for that question.

also  i mentioned there what i think is the best way to handle those resource files.
```
#pragma code_page(65001) //Make rc compiler happy with UTF-8 files
#include "en-US.rc"
#include "zh-CN.rc"
//etc...
```
this method reduces header bloat, only loading one copy for all included files.

NazmusLabs:
Yes I like the idea for the use of res.rc for this as well.

MouriNaruto:
@peterooch 
> i have only done and prefer to edit resource files manually, not in the resource editor of VS,
I edit the resource manually, but the compiler will return error if I use UTF-8 with BOM to save resource files.

Mouri

peterooch:
@MouriNaruto don't use BOM, BOM is evil and screws up UTF-8 files everywhere

MouriNaruto:
I followed your way just now.

Convert resource file to **UTF-8 without BOM** via Notepad++.


But the compiler **was not happy**. And I found It uses code page 936 defaultly (You canknow I use your way if you see my screenshots.).

MouriNaruto:
I also try to use command parameters in the resource compiler options to force it to use UTF-8 code page.
But the compiler returns "RC : fatal error RC1205: invalid code page".

MouriNaruto:
@peterooch 

Also I tried your idea ("only loading one copy for all included files") in UTF-16. Here is the code.
```
#include "res_en-US.rc"
#include "res_zh-CN.rc"
```

But the compiler returns "res_en-US.str(21): error RC2135: file not found: MENUITEM".

peterooch:
@MouriNaruto thank you for trying! I'll look later today at the files and hopefully I'll find a solution.

peterooch:
As PR #120 is more or less the same code wise now, I propose it's merging instead of this one, I think the files need to moved to a sub fo

craigwi:
@peterooch: I agree that we should focus on PR #120 and I agree with your suggestion to move the files to a sub-folder.

Issue #102, opened 2018-04-16
Winfile won't start since commit (dc4c149f1b0ba5e1a628b5e40557726640c710ef) under Windows 7


BaldwinTechnology:
![ms-api](https://user-images.githubusercontent.com/38319355/38810804-227276fa-4188-11e8-870f-a874916d50f3.png)

The program can't start, because api-ms-win-core-wow64-l1-1-0.dll is missing on your computer. Install the program again to solve this problem.

I have build the newest Commit (d024133a4816e42348ac7bc1a286fc037df90f0a): add GNUmakefile for build on mingw (#93)
and it doesn't start. But it builds.

The last commit which works is (af4b76e7c8d7a46d8d0f73572cfc3f4a7df990e3): Various improvements from PR #55.

I think is has something todo with the use of mincore.lib

System:
Windows7 with VisualStudio 2017 and
Windows SDK Version: 10.0.15063.0
Platform Toolset v141

Edit:

**Replace mincore.lib from linker settings with version.lib**

The online help for GetFileVersionInfo(), for example, has, for 4 years now, incorrectly listed mincore.lib as being required. It isn't; the correct one is version.lib. Perhaps you have a similar problem with misinformation.

Source: https://social.msdn.microsoft.com/Forums/vstudio/en-US/a28331ae-19a3-4a34-b3ba-1e8fd4430375/missing-apimswincore-dlls?forum=windowsgeneraldevelopmentissues

BaldwinTechnology:
Additional info found on:
https://social.msdn.microsoft.com/Forums/vstudio/en-US/a28331ae-19a3-4a34-b3ba-1e8fd4430375/missing-apimswincore-dlls?forum=windowsgeneraldevelopmentissues


As you said, the API Sets sound like they might be. The "umbrella" libs in API sets, MinCore.lib and MinCore_Downlevel.lib are provided in the Microsoft Windows Software Development Kit (SDK) that encompasses the API surface defined in API Sets plus additional APIs that are contained in well-layered system DLLs. So try include MinCore.lib and MinCore_Downlevel.lib maybe useful. I have search in my win7 and these are in SDK directory.  But notice that binaries that link to MinCore.lib or MinCore_Downlevel.lib **are not designed to work on Windows 7.**

No more Windows 7 support?

craigwi:
@matthewjustice, would you look at the above comment about mincore.lib?  Can we version.lib?

matthewjustice:
[...]

PR #103, opened 2018-04-16
Fix issue #102 Windows 7 issue with mincore.lib by using version.lib instead 

matthewjustice:
Addressed issue #102 by removing mincore.lib and replacing with version.lib. I was able to successfully build and test on Windows 10. I don't have a Windows 7 machine or VM available at the moment. It would be good if someone could verify the change on Windows 7 as well.

### DIFF
--- a/src/Winfile.vcxproj
+++ b/src/Winfile.vcxproj
@@ -109,7 +109,7 @@
       <AdditionalIncludeDirectories>.;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ResourceCompile>
     <Link>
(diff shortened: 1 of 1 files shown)


NazmusLabs:
Yup, I'm on it.

Biswa96:
Is that really need to mention the lib files in project file?Visual Studio automatically get those.

matthewjustice:
@Biswa96 - At build time the linker uses the project settings to determine which libs to use. A missing lib will cause linking to fail, since the linker won't be able to locate functions that are implemented elsewhere.  In this particular case, removing version.lib results in these errors:

`LNK2019	unresolved external symbol _GetFileVersionInfoW@16 ...`
`LNK2019	unresolved external symbol _VerQueryValueW@16 ...`

That said, if you have a suggestion for automating this somehow in Visual Studio I'd like to hear more.

Biswa96:
Oh, I forgot those Windows APIs. Sorry, for misleading...

PR #92, opened 2018-04-14
Enable visual styles.

Azarien:
This enables visual styles of WinXP and later.

Issue #88, opened 2018-04-12
What makes a good PR for WinFile?


craigwi:
Hi Folks,

First, I want to say thanks!  I am amazed and impressed by the number of people who care about this old tool.  Thanks to those who have contributed even if I screwed up the attribution on Github (see below).

In this "issue" I wanted to give you my thoughts on what makes a good PR and hear what you think.  

Thus far there have been several different kinds of PR:

1. syntactical changes like ^Z, spelling in comments, formatting, naming
2. results of automated checking tools such as putting parenthesis around macro arguments and fixing compiler warnings
3. visual changes such as the new icon and about box
4. changes for other platforms such as mingw (thanks @mattn)
5. functional changes (small or large) such as cold start window position issue or support for multiple search patterns

My thoughts: I am quite flexible on #1 and #2 and very much prefer them to submitted as separate PR from the other types.  This makes it easier for everyone to review and can focus our reviewing time/energy on the more substantial fixes.

Changes to improve the visuals (#3) are great as long as the spirit of the original look and feel is carried forward (more or less).  The updated icons and the Visual Styles change, which in the end makes sense, are good examples here.

I like the idea that this work can be used on lots of platforms, but as discussed in some issues the main line code should not suffer or be limited by them.  In some cases we probably need a separate branch.  I took the mingw changes because it seemed possible to do without impact on the main code.  Win31/NT3.5x should probably be in another branch. I'm not sure about WinXP yet.

Functional changes, #5, are the hardest type of PR for WinFile.  They are hard because the code is old and had many, many authors (before us) working on it for many years.  They are hard because there are lots of gotchas in the code, hard because some of the code is just too complex, and hard because it doesn't use modern languages or libraries.

To make a good PR which changes the functionality, that fits into the style of the code, and doesn't introduce other problems takes a lot of effort to get right.  Take @tig's simple one line change which was PR #46.  WinFile uses default coordinates in an odd (old) way and the fix for the real bug found was different than suggested.  

Questions:
* Can we keep PR focused on one type and one issue at a time?
* I've added a label "fixed differently" and just fix things based on the good ideas you have; would you rather I approach this differently?  If so, specifically how?
* When I make a change based on one of your PR and "fixed differently", how would you like the attribution to work?  It appears that I'm not doing it to everyone's satisfaction.
* Other comments?

Thanks,
Craig.

T8z5h3:
Craig i was going to write something about this today... 
[...]

PR #104, opened 2018-04-17
enable visual style on mingw

mattn:
related on #92

### DIFF
--- a/src/Winfile.vcxproj
+++ b/src/Winfile.vcxproj
@@ -278,6 +278,9 @@
     <ClCompile Include="winfile.c" />
     <ClCompile Include="wnetcaps.c" />
   </ItemGroup>
+  <ItemGroup>
+    <Manifest Include="winfile.exe.manifest" />
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
--- a/src/winfile.exe.manifest
+++ b/src/winfile.exe.manifest
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
+	<dependency>
+		<dependentAssembly>
+			<assemblyIdentity
+				type="Win32"
+				name="Microsoft.Windows.Common-Controls"
+				version="6.0.0.0"
+				processorArchitecture="*"
+				publicKeyToken="6595b64144ccf1df"
+				language="*"
+				/>
+		</dependentAssembly>
+	</dependency>
+</assembly>
--- a/src/res.rc
+++ b/src/res.rc
@@ -740,3 +740,15 @@ BEGIN
         VALUE "Translation", 1033, 1200
     END
 END
+
+#ifndef RT_MANIFEST
+#define RT_MANIFEST 24
+#endif
+
+#ifndef CREATEPROCESS_MANIFEST_RESOURCE_ID
+#define CREATEPROCESS_MANIFEST_RESOURCE_ID 1
+#endif
+
+#ifdef __MINGW32__
+CREATEPROCESS_MANIFEST_RESOURCE_ID RT_MANIFEST "winfile.exe.manifest"
+#endif
--- a/src/winfile.c
+++ b/src/winfile.c
@@ -13,14 +13,6 @@
 #include <commctrl.h>
 #include <shlobj.h>
 
-#pragma comment(linker, "\"/manifestdependency: \
-    type = 'Win32'                              \
-    name = 'Microsoft.Windows.Common-Controls'  \
-    version = '6.0.0.0'                         \
-    processorArchitecture = '*'                 \
-    publicKeyToken = '6595b64144ccf1df'         \
-    language = '*'                              \"")
-
 //
 // Overall Window structure
 //


NazmusLabs:
This enables visual styles on the about dialog box, correct?

ZanderBrown:
In theory it does it everywhere for all compilers

Not sure how this will affect the 'retro' branch but Visual Styles was always going to be an issue there

NazmusLabs:
@ZanderBrown That depends on whether we consider Windows XP support to be part of the retro branch or the master branch. If we keep retro branch for Windows 2000 or earlier, then there's no need to have anything related to the visual styles. If Windows XP, Server 2003 is part of the retro branch, well in that case, the MSDN article has a guide on how to have it fall back to version 5.

mattn:
I checked this PR can built on Visual Studio too.

craigwi:
I'm wondering if the split is at > WinXP (i.e., Win 2003 R2 and Vista).  

That is, we have a branch for pre-VistaWin7 and one for Vista/Win7+.

ZanderBrown:
So as far as I can tell this is the current proposal / state of play:
winfilever

ZanderBrown:
Sure, hadn't noticed @craigwi had said 2003 R2 not 2003
winfilever

NazmusLabs:
@ZanderBrown Okay, but the point is the same, is it not? Server 2003 is much, much, closer to Windows XP than is Windows Vista, no? As far as I know, the application platform's client APIs are identical to that of Windows XP whereas Windows Vista introduces new APIs on the client side. If I'm mistaken, do let me know.

P.S. I've finally recovered the archive of a genuine Server 2008 R2 VM I had and have successfully tested the visual styles compatibility in Windows 7 in high contrast. Result: it works, but there is a minor issue. If you switch from high contrast back to Aero theme, the toolbar icons become black boxes. An application restart fixes the issue. Another issue I noticed is that High DPI in Windows 7 and earlier does not work in this app when in High Contrast and classic themes. This is not a problem in Windows 10 and 8.x.

winfile in windows 7 server 2008 r2 aero

winfile in windows 7 server 2008 r2 high contrast

Notice that in the screenshots above, when in high contrast mode, the application is tiny.

ZanderBrown:
This is why we need to need to have this discussion!
winfilever

Seems a sensible place to draw the line

craigwi:
My POV is the same as @ZanderBrown expressed just above (master supports >=2003 R2 and >= Vista).

I am amazed at the support for earlier versions of Windows and will create a ver9x branch to support that (as requested in other issues).  I would like for the community to nominate one or two people to track / approve what's good for ver9x.  Who should that be?

Same for mingw.  Is that @mattn?

In case you don't know, I started at Microsoft before Windows 1.0!

ZanderBrown:
... and suddenly we are back to my original diagram :smiley: 

IIRC Vista is much closer to '03 than

ZanderBrown:
Okay so there was XP x64 & XP x64 Pro and it's the non-pro versions that were 'weird'
winfilever
My understanding of @NazmusLabs's latest proposal, complete with some previously omitted windows releases

NazmusLabs:
@zanderbrown indeed, there was. I was kind of hoping you would not bring up the "weird" version of XP 64-bit. That version of XP is best left forgotten. And that version of XP is not x64 even though it's 64-but. Crazy I know, but bear with me for a sec.

First of all, the 64-bit xp I was referring to was XP x64, was only in Pro editon, and was based on the same OS as Server 2003. I wasn't even considering the "other" 64-bit XP. That is a version of xp "that shall not be named". It is a software example of a dark past in the history of Intel, and a portal to an alternate, much darker, reality. Something that could have been, but didn't, and thank God for that.

Windows XP Pro x64 is exactly what you expect. It's literally a 64 version of Windows. It's the predecessor of all x64 version of Windows we have today, including Windows 10 x64. It works like how you expect, supporting both 32-bit and 64-bit apps. It has a Program Files (x86) folder, as do other x64 Windows. It has perfect SysWOW64, making all existing 32-bit apps compatible. So, supporting XP x64 is automatic. Merely supporting Windows Server 2003 will automatically support XP x64. In fact, if anyone were to run XP in a virtual machine today, they would almost always want to install XP x64, as it's better than XP 32-bit in every way. The only reason to use 32-but XP is if you want to use it on an old 32-bit processor ancient  hardware doesn't have 64-bit drivers.

Therefor, supporting XP Pro x64 is actually as important as supporting XP 32-bit, as anyone using XP in virtual machines will want to use the x64 version.

Now, that "other" XP. Right. That XP may not even launch the Winfile app, given you actually happen to find a still functionUS ancient PC capable of running that OS, one that was rare to get a hands on even in its heyday. No mainstream virtual machine software is, to my knowledge, capable of running Windows XP IA64. That OS was built for a reality that could have been, a reality we have thankfully, avoided.

Back in late 90s, we had left the 16-bit era. Windows 9x was the last OS that was a hybrid 32 bit and 16 bit OS, with Windows NT built as a pure 32-bit OS. By 2001, Windows XP was successful in porting all 9x features and DOS compatibility to the NT line. 32-bit was mainstream. Next stop: 64-bit. Intel had the "bright" idea to build a brand new 64-bit processor from the ground up. It was completely different from their current 32-bit architecture, which was based on thir 16-bit architecture which included the 286, 386, and 386, processors, hence the name x86. The new 64-bit Intel Processor was called Intel Itanium (IA64).

But there was a bad news. Since it was so different than the existing x86 processors, it meant the IA64 chips were dropping native support for all existing application. It was like moving from x86 to ARM64. Apps needed to be recompiled to IA64 to even run. You couldn't rely on just deploying an x86 app, expecting it to run perfectly, if at all. Windows XP IA64 did have a Wow layer, like how Windows 10 on ARM has, to emulate x86 apps. But apps would run slowly, would be buggy, and some wouldn't run at all. Had this actually happened, our computing industry would have been different. Compatibility with all retro software would be cut off. Winfile, or any other retro apps or games wouldn't just work with minimal changes in Windows 10 as they do today. That is, indeed, a dark reality. Macs may have continued using PowePC. The Xbox and PS4 consoles wouldn't be x86. But, thankfully, that is only a bad dream. Something that could have been, but hasn't happened. And the catalyst for that is AMD, which caused one of the biggest irony in the computer history.

AMD saw the stupidity of what Intel was trying to do. As it was struggling, it just realized that Intel was about to ditch the very platform that made them so powerful. AMD saw the golden opportunity. What if, rather than starting from scratch, they were able to figure out how to evolve the existing 32-bit x86 processor to 64-bit, as it was evolved from 16-bit. Sure, it's not as cutting edge as a brand new 21st century pure 64-bit processor. But the old x86 architecture was reliable, well understood, and most importantly, compatible. It could turn AMD around and allow it to be set for the next decade.

So they did. They created an evolved version of the x86 processor, that is essentially am extension of the 32-bit processor.

ZanderBrown:
Now @NazmusLabs has reminded us all of what we've tried to ignore with there epic post do we have some sense of agreement? @craigwi?

craigwi:
Wow.  Great write up @NazmusLabs!  I always like to make data-driven decisions :)

Regarding an earlier question, no I didn't work on WinFile before I copied the sources in 2007.  I was in the Office division from the beginning of my time at Microsoft in 1982 until late 1993 when I moved to Windows to ship 32bit COM/OLE in NT 3.51. Then I moved back to Office in late 1994, then to MSR in 1997, then back to Office in 2001 and then to Identity and Security in 2005 which is more or less where I am now (although on my fourth role -- my current role is what shows up on LinkedIn).

As for the primary question here, I agree with @NazmusLabs  and @ZanderBrown on the retro / master cutoff (red 2003 line in the most recent diagram).

src/winfile.exe.manifest
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
+	<assemblyIdentity
+		version="1.0.0.0"
+		processorArchitecture="*"
+		name="winfile.exe"
+		type="Win32"
+		/>
```
craigwi: We don't appear to need the "assemblyIdentity" section, do we?  Such a section introduces another version number to keep up to date.

src/Winfile.vcxproj
```diff
@@ -90,6 +90,12 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <IncludePath>$(IncludePath)</IncludePath>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <EmbedManifest>false</EmbedManifest>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <EmbedManifest>false</EmbedManifest>
+  </PropertyGroup>
```
craigwi: These should be removed (my preference) or the x86 versions made to match.

craigwi:
As to the PR, @mattn, I made a few comments on some details.  I'm in favor of taking this since the result is the same as the #pragma and it helps the mingw version.

@ZanderBrown, would you mind adding the most recent version of your diagram to PR #88 since that is now referenced by readme.md?

craigwi:
Also, I created the retro branch.  @MartinPayne and @NazmusLabs, go for it.  Perhaps PR #59 and perhaps others can be shifted to the retro branch.  It may go without saying, but we should keep the retro branch focused on essential differences.  I am open to changes to master that make that easier.

mattn:
@craigwi Thanks for your review. Updated patch.

craigwi:
Thanks.  One last question.  The change sets EmbedManifest to false for x64 (in vcxproj).  Would it be better to not set that value or set it the same for all versions?

mattn:
```
#ifdef __MINGW32__
CREATEPROCESS_MANIFEST_RESOURCE_ID RT_MANIFEST "winfile.exe.manifest"
#endif
```
If this part doesn't have ifdef, Visual Studio got an error (without EmbedManifest false). But now it is not required if it have __MINGW32__. So I removed EmbedManifest section.

craigwi:
Thanks.

PR #107, opened 2018-04-17
add depend target for GNUmakefile

mattn:
fix dependencies of .c and .h


ZanderBrown:
@NazmusLabs Yes & No, this is part of a greater effort allowing winfile to be compiled with GCC as well as MSVC & to run against Winelib as well as Win32. So it's true that when compiling with GCC you won't need the VC++ runtime but you will need the libgcc runtime unless it's linked statically. That make sense?

PR #112, opened 2018-04-18
modern c++

DBJDBJ:
Basically this is c++17 built with toolset 141  ..
wfgoto.cpp and BagOvalues.h have been almost rewritten.
All seems fine but the "Goto ..." feature does nothing at the moment.
Perhaps the original author might help in logic debugging :)

Thanks ...


DBJDBJ:
BagOValues.h :: I have deciphered the authors terminology and logic and am implementing the "proper" Retrieve logic and method.

Biswa96:
Windows SDK version 17110 is not currently available for normal users.

DBJDBJ:
@Biswa96, my mistake. What is the top SDK version accepted?
I assume toolset 141 is OK.

Biswa96:
It's current released version is 10.0.16299.91. See this link: https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk 
But i don't know what SDK version need for this project to be compatible old WinXP and Vista.

DBJDBJ:
Making a decision to go "UWP only" will clarify the roadmap. Among other things.

ZanderBrown:
> a decision to go "UWP only"

Seems very unlikely

Particularly as there is an effort to backport as well as keep moving forwards how does this help us?

*If it ain't broke don't fix it?*

DBJDBJ:
that little cpp part was more than broken actually ...  I spent a day fixing it to the poing when it does nothing LOL ... just kidding .. need to understand the logic behind .. it is a small part of WinFix ..

ZanderBrown:
That's about making it available on the Store, which has little to do with UWP

NazmusLabs:
@DBJDBJ yes, this is one of the most common misunderstanding that dates to how things worked in Windows 8. Universal Windows Platform (UWP) apps doesn't imply Windows Store version and vice versa. The Windows Store can deploy apps of various platforms, including Win32, UWP, Silverlight 8.1 (for legacy Windows Phone apps), and the legacy Xbox One C++ games that are not built using UWP.

The converse is also true. UWP apps DO NOT need to come from the Windows Store. A user can simply download a signed UWP app package from the web, double-click to launch a standalone installer which will install the UWP on their computer. Furthermore, third party launchers and stores can also deply UWP apps. This means, ABSOLUTELY YES, Steam can distribute UWP games. A living example of this is Adobe Experience Design CC. This is a UWP app; however, it is not available form the Windows Store. It is distributed through the Adobe Creative Cloud desktop app and installed through it, like other CC apps, such as Photoshop CC and Premiere Pro CC.

Thanks @ZanderBrown for also clarifying this. :)

DBJDBJ:
Ok then Windows Store it is, I assume? I have no preferences to yes or no. It is just that such a decision will clarify the Road Map for WinFile considerably.

While I think it this project philosophy should be very simple: Just leave it as it is :)
(of course fix the bugs that inhibit the functionality, like help etc ...)

This little c++ part of it I am dealing with perhaps should not be in at all, but ok lets fix it ..

DBJDBJ:
Thanks for the discussion guys. I am closing this pull request. I need more time to understand better the logic behind the "Goto..." feature ...

Issue #111, opened 2018-04-18
No Toolbar-Buttons since (49597b8c423b8e8d16ee252adc1e6e8f69e8c1ac)


BaldwinTechnology:
Hi,

the toolbar **buttons** are gone with the checkin (49597b8c423b8e8d16ee252adc1e6e8f69e8c1ac) 

Squashed commit of the following: 
16. April 2018 10:44:13

Only the pictures are left. Is this a Win7 only behavior or a preparation for visual styles?

The following lines are addes in winfile.c

`#pragma comment(linker, "\"/manifestdependency: \`
`    type = 'Win32'                              \`
`    name = 'Microsoft.Windows.Common-Controls'  \`
`    version = '6.0.0.0'                         \`
`    processorArchitecture = '*'                 \`
`    publicKeyToken = '6595b64144ccf1df'         \`
`    language = '*'                              \"")`

Greetings
Daniel

thecatkitty:
It is not a bug. Moreover, it is an **intended part** of using the visual styles.

BaldwinTechnology:
I have supposed something.
But how can I get the old style back?

thecatkitty:
I am not sure, but perhaps setting __Disable visual themes__ in winfile.exe compatibility properties will help?

ZanderBrown:
@thecatkitty that may help but I seem to remember that disables Aero as well?

Having this as an option should probably be investigated

thecatkitty:
@ZanderBrown Windows _of course_ [enables us to do that](https://msdn.microsoft.com/en-us/library/windows/desktop/bb773175(v=vs.85).aspx#turnoff). We might add such option to the menu (it starts to seem like the _Preferences_ dialog could become a good thing to add) but it would create a dependency on `uxtheme.dll`... Shall we use LoadLibrary then?

Biswa96:
First check if that API works for WinXP or Vista.

ZanderBrown:
Yes LoadLibrary would seem the way to go, I think for now the Options menu is fine but maybe we should have a separate dialog for all the additions over the original WinFile features

NazmusLabs:
This issue has been now fixed in the latest commit. I think this can be closed.

PR #113, opened 2018-04-18
Make it possible to disable Visual Styles for the toolbar (#111)

thecatkitty:
I've added a new item into the _Options_ menu which, by using **SetWindowsTheme** from dynamically loaded `UxTheme.dll` (so it should work if we tried to backport to 2000 or older), disables or enables the usage of Visual Styles in the toolbar.

Selected setting is, of course, saved in the INI file.


NazmusLabs:
Neat option to have it dynamically switch; however, I am failing to find a scenario in which you would want to disable the visual styles. Perhaps I'm missing something, and, thus, would love to know your opinion on the value of this.

thecatkitty:
@NazmusLabs, as it seems in #111, there are people who'd like to have the _old-school_ look of the Winfile without messing with the code. Giving them such an option out-of-the-box would be, in my opinion, a good move.

Is it practical? Certainly not. Is it a nice thing to have? Why not.

NazmusLabs:
@thecatkitty Ah, I see

Anixx:
Why toolbar-only then? Would not it be better to have similar application-wide option?

Also, I definitely do not understand people who run a visual style they do not like.

thecatkitty:
@Anixx, of course we _could_ do that, but that would mean one of two things:
1. Disabling Visual Styles on per-every-dialog-and-control basis, which surely is possible but it would be kinda overdoing and would provide just a few minor noticeable visual changes. (Just the  appearance of buttons and boxes in dialogs, really) OR
2. Disabling styling on per-process basis, thus loosing also window decorations, which would make it look completely unnatural. And I am not sure, if that wouldn't also make our program incompatible with Windows 10 incoming _Sets_ feature.

About Your second statement, I am not sure if we're talking about the same thing. Since Windows 8 changing or disabling Visual Styles per-system is a very custom and really unsupported thing. Moreover, in the Windows 10 disabling Visual Styles (which is in fact messing with the handles of  the Win32 subsystem) makes a half of the interface unusable.

NazmusLabs:
@thecatkitty to my knowledge, it doesn't actually disable Window decorations on Windows 10 or 8. That's because when Desktop Window Manager (Aero) is enabled, it handles the Window boarders, including the titlebar. So on Vista and 7, when Aero is enabled, it doesn't affect the Window boarder. On XP and with Aero disabled on Vista/7, you would get the classic boarders.

Thing is, you can't disable Aero (DWM) on Windows 8/10, so it should work with sets and would look fine.


Correct me if I'm wrong.

thecatkitty:
@NazmusLabs, when I disabled Visual Styles of hwndFrame on Windows 10 it gave me the "classic" window borders an title bar too.

But I'll keep checking.

Anixx:
@NazmusLabs  @thecatkitty if the styles are disabled per application (such as when runbning the ypplication wil manifest which does not support themes or when using compatibility layer), the window decorations are not affected (would be styled as systemwide). 
Only if to disable the styles systemwide (using handle or whatever), the decorations become unstyled (even when DWM still works as on my screenshot of Win8.1 below).
clipboard_20170911

Anixx:
@thecatkitty Moreover, in the Windows 10 disabling Visual Styles (which is in fact messing with the handles of the Win32 subsystem) makes a half of the interface unusable.

This depends on when you are disabling the themes: before or after the UWP/Modern engine starts (it is in separate process in Win10). If you disable the themes after it started, you have no styles systemwide (for programs started after disabling), but UWP/Modern/Metro still working.

thecatkitty:
Unfortunately I am not able to attach a screenshot, because I'm being AFK right now, but when I called SetWindowTheme on hwndFrame, I got a window with unstyled decorations on Windows 10 with every _beautifier_ enabled. I seriously do not consider that a feature.

If someone has their Visual Styles disabled systemwide, they'll get everything unstyled, of course. But I think, that if someone wants to get an oldschool look of the Manager in a styled environment, I bet they mean having unstyled *client area*, not unstyled window.

Anixx:
> But I think, that if someone wants to get an oldschool look of the Manager in a styled environment, I bet they mean having unstyled client area, not unstyled window.

I cannot guess what such people may want because if I dislike a style, I change it systemwide.

NazmusLabs:
@Anixx yeah, that's what I was thinking when I write my initial reply. Typically, you don't see apps have an option to enable or disable the visual style, as, traditionally, it was expected that users would pick their favorite style and the app would just adapt. That was the intention, after all. The visual styles API just sets the app's UI controls to use whatever the current OS is using.

thecatkitty:
@Anixx, that's how it looks like when You disable Visual Styles for hwndFrame in Windows 10 with Themes enabled (everything default):

As we can see, window decorations are unstyled and window metrics aren't perfect. That's why I've chosen to affect the rendering of the toolbar only.

I just believe (basing on the evidence in the linked issue) that there are people who'd like to have good old boxed toolbar buttons, that's all.

ZanderBrown:
Wow that's a weird sight, wonder why the border is so large (especially when Win10 has them so thin, if at all, by default)

I think we can safely say that's not what people want/expect

NazmusLabs:
@thecatkitty oh, wow that's a rare sight to see. And yeah...I don't think that's what people would want. It just doesn't look right. And I stand corrected about the boarder thing not being applicable with Windows 10's DWM enabled.

Anixx:
Definitely there are methods to untheme only client area. For instance, if one starts Outlook Express under Win8.1 with "no themes" compat flag, only the client area is unthemed.

thecatkitty:
@Anixx, I wouldn't be very surprised if they just disabled styling on per-control basis.

NazmusLabs:
I think this should be an option that should exist in INI file rather than on the menu. Most toggle options in the menu are those that we might want to change often and bees quick access, such as setting sort order and view nice. I think most people will keep the visual styles on, and those who want to turn it off, will do so once. It is unlikely people will be turning it on and off frequently. So this kind of "set and" forget config would be better off tucked away in a preferences Window or an INI file.

Secondly, it appears only enthusiasts or troubleshooters will want to tinker with this option. So it should be left out of sight for most users anyway.

Anyone disagree?

thecatkitty:
@NazmusLabs I think it is a good idea. It isn't going to be a setting to change very often.

Creating an _Additional preferences_ dialog would be a good solution.

ZanderBrown:
@thecatkitty :+1: but if we don't want to do that the system menu could be a good place to put it

craigwi:
I think @NazmusLabs suggestion makes sense (winfile.ini only option).  That would meaning removing the changes to the menus.

Also, @thecatkitty, please separate out any formating / spacing changes so that they don't get included in a PR with functional changes (e.g., changes at top of res.rc).

You will also need to resolve the conflicts with winfile.h after res.h was separated out.

thecatkitty:
Okay, I've resolved conflicts and removed every Visual Style disabling-related interface element.

@craigwi, so there's going to be some kind of dialog box for _Preferences_ or it'll be just a manually-edited INI file?

craigwi:
Changes look great.  I expect we will have a Preferences dialog in the future.

PR #114, opened 2018-04-18
App Version Updated to Version 1804

NazmusLabs:
The application version has been updated from version 10.0.0.0 to version 10.0.1804.0. I've not updated the major or minor version; so it's still version 10.0 as of now. The 1804 signifies that it is the April 2018 code.

file manager version 1804

Since there is no automated versioning system implimented yet, this allows us to at least easily update the version without worrying about whether to increment the major or minor version.

Note: I originally wanted to go with something like 180418, (YYMMDD) to include the day. This way, whoever commits a code, they would just update the version number to reflect the date. Then it'd be super easy to tell which "build" one is runnign for troubleshooting reasons. But I went with just YYMM (1804) because we'd only need to increment the version at least once a month.

But I'm interested to hear what you think of how versioning should be done.

### DIFF
--- a/src/res.rc
+++ b/src/res.rc
@@ -700,10 +700,10 @@ END
 #define VER_PRODUCTNAME_STR         "File Manager"
 #define VER_LEGALCOPYRIGHT_STR      "Copyright (c) Microsoft Corporation. All rights reserved."
 
-#define VER_FILEVERSION             10,0,0,0
-#define VER_FILEVERSION_STR         "10.0.0.0\0"
-#define VER_PRODUCTVERSION          10,0,0,0
-#define VER_PRODUCTVERSION_STR      "10.0.0.0\0"
+#define VER_FILEVERSION             10,0,1804,0
+#define VER_FILEVERSION_STR         "10.0.1804.0\0"
+#define VER_PRODUCTVERSION          10,0,1804,0
+#define VER_PRODUCTVERSION_STR      "10.0.1804.0\0"
 
 #ifndef DEBUG
 #define VER_DEBUG                   0


thecatkitty:
I appreciate the idea of date-based versioning scheme a lot. <3

Will the granularity of year + month be enough? Probably, but maybe we could just use one additional digit in _release_ version number to indicate week of the month, eg. today would be 10.0.18044 (or 10.0.18043, if we're counting from 0). We can't be sure if we're not going to release two versions on the same month.

As we're talking about the fourth (_build_) version number, we could use it to indicate the number of commits in *master* branch since the previous release, but that would make the repo manager/owner increment it on every accepted pull request. (It would make 10.0.18044.77 then). But I'm afraid that I am overcomplicating right now...

craigwi:
We definitely need a versioning scheme.  Hard coding it to res.rc does not seem like the right idea.

I'm still interested in appveyor.com; that should provide us a lot of flexibility.

NazmusLabs:
@craigwi yes I agree. I intended this to be an interim/temporary thing. Currently, version 10.0.0.0 is hard coded. So, if something is going to be hard coded, I thought it would be something other than 10.0.0.0 to distinguish it from release. Perhaps others beg to differ?

NazmusLabs:
And I think using appveyor thing is brilliant. Played around with it. It can be set up pretty easily, it appears.

ZanderBrown:
Perhaps version should also include m or r to signify a master or retro build

thecatkitty:
@ZanderBrown Weren't we going to make the _retro_ branch number 9.0?

ZanderBrown:
@thecatkitty , yes we were, ignore me :slightly_smiling_face:

NazmusLabs:
@thecatkitty I definitely wanted to do something with the very last revision number, and number if PR seems like an awesome idea. However, there's is one teeny tiny major problem.

It's something I cannot comprehend at all, and would LOVE it if someone knew why this problem I'm about to describe exists.

Windows Store, for whatever reason, no longer accepts Windows 10 packages that have a version number in the fifth (aka revision) section to be anything other than zero. I don't understand this at all. I had to redo my entire version in for one of my apps that I tried to submit to the store because of this restriction.

So, for example, a package of version, say, 3.2.38.57 will throw an error if you try to upload it to the store. You must increase its version number to something like 3.2.39.0. Essentially, the last number must be zero.

This wasn't the case in Windows 8.x. And when you install or update apps in the store, you'll notice that all Windows 10 apps have a version with the last number being zero. And if you install an app that still hadn't yet been upgraded to target Windows 8, you can see that the version number either in the Store or the settings app likely has a non-zero value in its last number.

This means the last number is essentially useless as a versioning scheme. A non-zero value can only signify beta releases and be a package incapable of being uploaded to the store

Anyine have any idea for the reasoning behind this artificial restriction?

T8z5h3:
@NazmusLabs is the .exe updated on get hub?

PR #117, opened 2018-04-19
Updated Readme file to include link to #88

NazmusLabs:
The issue thread #88 offers some great points on what makes a good pull request. However, it is buried under a lot of different issues and can be difficult to reach, especially for new contributors who might not know this discussion exists. As such, I felt it necessary to add a direct link to the thread under the "Contributing" section of the readme file.


craigwi:
Thanks @NazmusLabs.  I have a backlog of great comments to review.

Biswa96:
Isn't it be better if you place the URLs in separate section? Like following example:

```
Here is the [link](https://example.com). See this.
```
After placing the  URL in separate section, this will be:

```
Here is the [link]. See tis.

[link]: https://example.com
```

Biswa96:
Isn't it be better if you place the URLs in separate section? Like following example:

```
Here is the [link](https://example.com). See this.
```
After placing the  URL in separate section, this will be:

```
Here is the [link]. See this.

[link]: https://example.com
```

NazmusLabs:
@biswa96 sure. I was debating this myself. Decided to not have a redundant link on my commit, but I'm equally favorable to the other approach you noted.

PR #118, opened 2018-04-20
Goto Directory implementation

DBJDBJ:
All seems fine. 
wfgoto.cpp and BagOValue.h completely rewritten. existing functions/interfaces  same as before.
Goto Dir made to work on the main thread. Seems working OK.
As far as Task Manager reports, Private Memory set seems stable and not rising.


DBJDBJ:
x64  build reports quite a few warnings on type  mismatches ... inside the C code ...

DBJDBJ:
Hello @craigwi , Perhaps a better implementation :) meaning using modern C++. No pointers, no even smart pointers, bagOvalues is just a thin front to std::multimap, etc. 

1. This PR has the functionality crippled and I apologize for submitting it too early. Version on "my side" now seems to work ok + it indexes the current mdi child selected drive not just the C:

2. I made indexing not to start before the goto dialog is up.  So the current drive is re-indexed each time goto dialog pops up. I do realize on smaller machines this might be a problem.

3. Conceptually "Goto Drive" of Win File represents something of a very  limited functionality of the current File Explorer Search.  I do realize we can not use that API if we are to serve Win versions way back to W98? or Even XP which had/has different Search API.  Using WIN Search API seems thus out of question.

4. I think the feasible course of action is:

4.A  -- use first just new bag_of_values. Test it. 
4.B  -- on the parallel track formalize the scope and thus requirements for the improved "Goto Folder" (goto folder dialogue improvements and a such). For example one architecture of the solution might be to have in memory cache for each drive. And then have it updated on folder tree changes events.
4.C -- develop and test the new logic and front of the new Goto File.

5. Instead of me lambasting you with PR's I might humbly request testing [https://github.com/DBJDBJ/winfile](https://github.com/DBJDBJ/winfile)
It is (for me) very difficult to find a way through winfile code.

6. If and then fully developed "Goto Folder 2" will certainly be one non trivial "subsystem" of the Win File. Thus we/you might decide to deliver just a limited functionality. After the phase 4.B we should know ... Certainly it will not be simple :)

7. If you agree I will close this PR and deliver just a modern bag_o_values in the next PR as per point 4. Leaving everything else untouched. After point 5 and my bugs sorted we might have all in order and I can deliver the next PR .

I hope I am not too confusing :)

DBJDBJ:
off I go closing this PR and preparing much simpler PR

PR #120, opened 2018-04-21
Add Chinese Simplified translation

MouriNaruto:
- Save resource file as UTF16 to resolve character encoding problem. 
- Split Resources for multilanguage support. 
- Add Chinese Simplified translation (Based on NT4's translation, but with some modifications.). 

Mouri


peterooch:
2 things to say...
1. This similar to PR #101 (though different scheme for the files)
2. with UTF-16 github doesn't show the new files at all... UTF-8 doesn't work with simplified chinese?

MouriNaruto:
@peterooch 

1. I didn't notice your PR until I made this PR. And this PR add the Chinese Simplified translation.
2. Because Visual Studio defaults to saving RC files as UTF-16. So I followed the Visual Studio.

Mouri

peterooch:
@MouriNaruto i would recommend to change how the resource files are include,
I think its better off to #include them in res.rc so to avoid needless inclusion of the same header files.
for example:
```
#include res_1033.rc
#include res_2052.rc
```
and also using `res_enUS.rc` and `res_zhCN.rc` would be more clear than 4 digit codes

MouriNaruto:
@peterooch 
I renamed the resource files.

But I don‘t want to include res.rc because of the resource ID conflict.

peterooch:
@MouriNaruto 
Remove in res.rc
`#if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_*)`
and the accompanying `#endif` at the end of the file

in res_en-US.rc and res_zh-CN.rc remove all the pre-processor lines until the `LANGUAGE` statement and removed the `#endif` at the end of file
then `#include` both above mentioned files in the bottom of res.rc and voila, it compile with both languages.
```
#pragma code_page(65001)
#include "res_en-US.rc"
#include "res_zh-CN.rc"
```
The `#pragma` directive makes the resource compiler to handle UTF-8 characters properly

There would not be ID conflicts because you added the necessary `LANGUAGE` statement
(I did a half baked Hebrew translation and it worked like that)

Please mind this is not criticism of you pull request at all, just my recommendations.

MouriNaruto:
@peterooch 

I tried but I found if I used your solution to use UTF-8, Visual Studio Resource Compiler will report an unknown error.
PS: In one or two years ago, I tried to use UTF-8 to save RC file in my open source project but failed. So I known save RC file as ASCII or UTF-16 can make Resource Compiler work successfully. I think if you want to use UTF-8, use JSON or XML will be better.

Mouri

MouriNaruto:
@craigwi Do you have some suggestions about this PR?

kent7301:
Здравствуйте!


IOS

mattn:
@MouriNaruto Please fix conflict. thanks. If this will be merged, I'll add japanese translation.

peterooch:
@MouriNaruto please try the files attached in the zip file, they are the same files with my proposed changes,
only thing required is to set "res_enUS.rc" and "res_zhCN.rc" properties in the solution file from "Resource compiler" to "Does not participate in build".
[pr_120_fix.zip](https://github.com/Microsoft/winfile/files/1945465/pr_120_fix.zip)

MouriNaruto:
@peterooch Thank you very much. I also learned something and used it on my own projects.

craigwi:
Looks like a simple solution; reviewing now.

craigwi:
@MouriNaruto, how can I test this without changing the default UI language?  I have the required language pack installed.

MouriNaruto:
@craigwi 

I've never tried to test this without changing the default UI language.

Maybe this can help you.
https://msdn.microsoft.com/en-us/library/cc194810.aspx

craigwi:
I already tried that (calling SetThreadLocale()).  FindResourceEx does not find the cancel dialog:

   LCID chinese = MAKESORTLCID(LANG_CHINESE, SORT_CHINESE_PRC, SUBLANG_CHINESE_SIMPLIFIED);
   HRSRC hrsrc = FindResourceEx(hInst, RT_DIALOG, MAKEINTRESOURCE(CANCELDLG), chinese);

hrsrc is NULL.  Other ideas anyone?  What am I missing?

peterooch:
@craigwi one idea is to comment out the en-US resource include, so it would compile only with the chinese resources

thecatkitty:
@craigwi You are not missing anything. It just seems that `SetThreadLocale` doesn't work as expected. But I've managed to get Chinese UI by calling the following:
```c
SetThreadUILanguage(MAKELANGID(LANG_CHINESE, SUBLANG_CHINESE_SIMPLIFIED));

craigwi:
I had called SetThreadUILanguage too, but didn't use the correct parameter (was missing the sub language).  Thans @thecatkitty.  It works now!

craigwi:
Awesome.  I commited a change (of my own making 😀) which adds a new option to winfile.ini: UILanaguage.  To test @MouriNaruto's changes, I set UILanguage=zh-CN and restarted.

@MouriNaruto, I have a few small requests for your change:

1. the code in suggest.c related to suggest.db has been disabled when you didn't include suggest.db.  Please look into reenabling that.

2. the strings in res_en-US.rc between JAPANBEGIN and JAPANEND have changed.  @mattn, are you going to handle that?

3. let's put the en-US, zh-CN and all other files in a subdirectory, say lang.  So res_en-US.rc would move to lang\res_en-US.rc as would all of the other translations.  I'm hoping we have a lot of them and we can keep them together.

MouriNaruto:
@craigwi 

I have finished No.1 and No.3 of your requests.
But I can't do No.2 of your requests because I have not learned Japanese. You need @mattn ‘s help.

peterooch:
Maybe put them in a separate file? I doubt they are going to be changed at all in each translation.

MouriNaruto:
@mattn
In my opinion, I think we can drop JAPANBEGIN/JAPANEND in res_en-US.rc in this pull request.
But if you want, you can move them to res_ja-JP.rc when you add the Japanese translation.

thecatkitty:
I wonder why were the Japanese strings there in the first place.

Perhaps it had something to do with weird floppy disk drive types in Japan?

craigwi:
Thanks @MouriNaruto.  @mattn can you take a look at all of the JAPANBEGIN/JAPANEND and the use of bJapan, bJAPAN and bKOREAJAPAN?  Seems like we can just delete all of that code.

MouriNaruto:
Thanks

Issue #115, opened 2018-04-19
Customizable text editor


mattn:
I hope to customize text editor to use my favorite. Add new dialog?

thecatkitty:
I think that that's exactly the purpose of the Windows file type association mechanism.

To be honest, such dialog does already exist in the File Manager...

mattn:
AFAIK, people who don't like to modify registry or file type association exists. (include me)

mattn:
For example, how about to read winfile.ini ?

craigwi:
While I suggested essentially what @thecatkitty said, I like the idea of a Winfile.ini entry (not a dialog box in the UI) to specify the notepad-link editor.

NazmusLabs:
@creigwi @thecatkitty I'll give you a golden example. Say you happen across an HTML file. Double clicking it in file manager will open your web browser. What if you wanted to see the source in notepad. Forget notepad++ or VS code. We're just taking notepad.exe. Same goes for, say, a docx file. Double clicking it opens Microsoft Word. But you might wish to right click and just view the XML in notepad. We might want to add a simple "Open in Notepad" to simply open the file in notepad. Configuration options to set different text editors can be added on later.

Funny enough, I was trying to troubleshoot my midi driver on my Windows 3.1 in DOSBox and tried to just see the contents of a file without associating that file type with notepad system wide. I simply wanted to be able to right select the file and click edit to have an one time instance opening in notepad. I eventually had to open notepad and use the open dialog box to open that file. But now, I really get the point of this feature request.

Sometimes you just want to view a particular file in plain text just occasionally, and changing file associates don't usually make sense in those cases.

NazmusLabs:
@petrooch wow, since when did you have a full blown preferences dialog box? Or did you just create this now?

A preference window would be a wonderful addition for all future configuration options as well as consolidating existing, discrete, configuration windows and menu items.

peterooch:
its from 7-zip... just meant as an idea for what it could be,
i think i can build a dlg similar in fashion, but somebody else will have to figure the method for it

NazmusLabs:
So reading the comments above, I, too, like the idea of having a simple ini file, but I'd go further in suggesting that we eventually move all the existing configuration options to the ini file, making it the one, universal, comfig store for the application. And if and when a single preferences GUI is created, it would be based on the ini file, with settings being reflected on the GUI preferences being reflected in the INI and vice versa. And perhaps some less commonly used settings stays exclusively to the ini file.

[...]

PR #121, opened 2018-04-21
Additional Prefrences dialog

peterooch:
Added a dialog that enables a selection of text editor and view other than notepad,
currently works only if manually specifiying a path to the editor.
The "..." buttons throws an access error.

Saves the path in the INI file

Includes fallback to notepad if something bad happens

Made by inspiration from Issue #115 

(also uses wcs* functions instead of the tcs* equivalents)


src/res.h
```diff
@@ -144,11 +144,13 @@
 
 #define IDM_TOOLBARCUST     512
 
+#define IDM_EDITOR          513
```
craigwi: No need to renumber existing entries here.  Just add later in the file.

src/wfcomman.c
```diff
@@ -690,31 +690,24 @@ OpenOrEditSelection(HWND hwndActive, BOOL fEdit)
...
-            if (!PathFileExists(szToRun))
-            {
-                cchEnv = 0;
-            }
-          }
+      {   
 
-          if (cchEnv == 0)
-          {
-              // NOTE: assume system directory and "\\notepad.exe" never exceed MAXPATHLEN
```
craigwi: Carry forward this comment.

src/wfdlgs.c
```diff
@@ -698,6 +698,72 @@ ConfirmDlgProc(HWND hDlg, UINT wMsg, WPARAM wParam, LPARAM lParam)
...
+            break;
+        
+        case WM_COMMAND:
+            switch (GET_WM_COMMAND_ID(wParam, lParam))
+            {
+                case IDD_HELP:
+                    goto DoHelp;
+
+                case IDC_EDITOR:
+                    GetOpenFileName(&ofn);
```
craigwi: This should use the LoadComdlg() pattern elsewhere in the code.

src/winfile.h
```diff
@@ -962,7 +963,7 @@ Extern DWORD (*lpfnGetOpenFileNameW)(LPOPENFILENAMEW);
 #define COMDLG_GetOpenFileNameW           "GetOpenFileNameW"
 
 #define ChooseFontW         (*lpfnChooseFontW)
-#define GetOpenFileNameW    (*lpfnGetOpenFileNameW)
+//#define GetOpenFileNameW    (*lpfnGetOpenFileNameW)
```
craigwi: You should not need to remove this.

peterooch:
prefdlg_show 2

Added more stuff... is it too much feature creeping for a PR or it would be a welcome addition/change?

thecatkitty:
@peterooch, I know that it's not an answer for your question, but wouldn't it make more sense if the string "Chinese (Simplified)" were written in Chinese?

peterooch:
@thecatkitty, Its in a placeholder string table, just to see if it works, the current branch doesn't have the a9fe81f code even, but i had to have something to differentiate from English. I also would prefer "Local Name (English Name)" or the other way around for the list values.

craigwi:
I like the direction.  We should move the relatively new option "Create Goto Index on Launch".  I would also be ok with moving all of the other binary options (the last four on the current menu).

ZanderBrown:
I think it would be better for the original options to remain where they are but we could add them to the dialog

tl;dr copy not move

craigwi:
See commit comments for the few tweaks I made.  Thanks!

PR #122, opened 2018-04-21
Resizable dialogs for common dialogs

malxau:
This makes some of the top level dialogs resizable.  The resizing support uses markup in the dialog resources and is fairly straightforward to use.  Note this doesn't work on Win9x but support for that is no longer present in winfile.  Some dialogs can't be resizable that should be; I can't seem to make fonts work for example, which is a custom template on the common dialog library.  Things like progress or confirm dialogs don't seem to add value by being resizable.

The per-dialog resize states are conceptually statics.  They're just blobs that are allocated and destroyed along with the dialog box, and it's safe to overload them so long as two dialogs can't exist concurrently trying to use the same context.


NazmusLabs:
At first I wasn't sure what you were trying to achieve. So, I downloaded your commit and built the solution.

After having experienced with your changes, I can definitely see the value of this. The dialog boxes can be widened to increase the view is the text boxes that exists in many of these windows. This allows one to be able to see more of the text without having the text scroll, which is especially helpful for many of the text boxes which allows you to type in a directory path.

craigwi:
Every day I learn new things -- like this interesting idea!  I need to review in more detail to comment.

craigwi:
Hi @malxau, given that master now has 2 soon to be 5 languages, have you given some thought about how to simplify the feature and not require every dialog field to be annotated?

malxau:
I re-merged this with the current localized dialog resources.  Please let me know if this is something to proceed with or not.  If it's not something to proceed with I can go do something else.

craigwi:
Definitely still interesting.  Given there are more and more things to get right, I built a simple resource verification tool that makes sure all of the languages have the same resource identifiers with the same attributes as necessary.

NazmusLabs:
Have you, @malxau, tested it working on all currently available languages? If not, I can do that. I'd just need to pull from your latest commit and rebuilt. I currently still have your older code, but I don't mind testing your latest code if it helps you.

malxau:
@NazmusLabs That'd be greatly appreciated.  Out of interest, how does this work? How do you test a specific language?

When testing, please look at pl-PL's SELECTDLG and RUNDLG.  The first appears to be larger than the en-US one and the second has a mighty long string that I'm expecting to collide with the buttons if they move.

Also, I haven't looked at he-IL yet.  This one is a stress test for this code because everything is right to left.  The code is designed to handle it and I have simple tests that demonstrate it working, but it is a little different.

malxau:
As a concrete illustration of the long text above, look at how "Run as Admin" is positioned in English vs. German.  If reposition information is with the control, this is straightforward; if it is stored independent of language layout, it will be buggy.

run_en
run_gr

malxau:
It looks like `VerifyResources` has a bug in processing these resources.  In `DecodeDlgItem` it treats cbCreate as being the total number of bytes to advance (including the cbCreate field itself.)  I know this is exactly what MSDN says it means, but after looking at it, I'm fairly sure MSDN is wrong: the cbCreate field needs to be accounted for unconditionally, and the number of bytes in it are additional to that.  This hasn't mattered before resizeable dialogs because no dialog item had additional creation data, so only the first (and correct) branch is taken.  With these changes, the offsets of subsequent controls are not interpreted correctly, and the code will complain about all kinds of spurious things, since it will interpret localized strings as control IDs, etc.

Somewhat unrelated, but this tool is flagging issues in master now which appear legitimate:
```
RT_CURSOR
RT_BITMAP
RT_ICON
RT_MENU
RT_DIALOG
Warning: dialog Confirmation item #6, id 215 for language pl-PL does have the required style 0x50010003
Error: dialog Preferences for language de-DE does not have the required number of items 12
Warning: dialog Moving... item #2, id 108 for language ja-JP does have the required style 0x5000008b
RT_STRING
Error: string table block usage mismatch for language tr-TR block 15
Error: missing lanaguage zh-CN for resource type RT_STRING, name 126
Error: string table block usage mismatch for language zh-CN block 126
RT_ACCELERATOR
RT_GROUP_CURSOR
RT_GROUP_ICON
RT_VERSION
RT_MANIFEST
```

schinagl:
Just donwloaded the branch and I also find this neat and a valuable extension to Winfile.
What are the problems? 
Can we step by step work them off?

malxau-msft:
I updated this to current.

In terms of the problems, everything is covered in this issue.  I think it's just a large change to review.  There's obviously a desire to make something that applies to all languages, although I'm skeptical that doing so works reliably if dialog layouts can differ.

If you give this a serious review and approve it, then I think it can be merged.

schinagl:
Took it over into my daily work branch and give a test for a week

schinagl:
Had this running on daily basis for almost 2 weeks and it worked flawlessly!
So you have my approval! 
Please merge it.

craigwims:
reviewing now.

src/Resize.c
```diff
@@ -0,0 +1,954 @@
+
```
craigwims, line 1: Need to add the standard file headers as on all other files.

craigwims:
As we discussed, this PR should be broken into parts, but my suggestion is simpler than the one we talked about.  Separate out the change in VerifyResources.cpp as it is completely independent.  Also separate out the changes to res*.rc since they are unrelated to the resizable dialogs.  Then we'll tackle the main change.

On the main change, the use of global state for ppStorage of ResizeDialogProc is old school :).  Are there no other options for saving this?  GWLP_USERDATA would be ideal, but it is not generally available and it would be non-trivial to use it.

schinagl:
How to precede with this one? Looks like not much is missing? So could it be merged?

Issue #54, opened 2018-04-10
The color of the folder edge and tree is grey instead of black


Anixx:
The color of the folder edge, "+" signs and the tree branches is grey instead of black
![clipboard_20180410](https://user-images.githubusercontent.com/1271736/38554208-30cfa23a-3cb1-11e8-98b6-98aefdf5d01b.png)

Original:
http://storage4.static.itmages.ru/i/17/0919/h_1505825396_3777912_92ff09bce8.png

ttsirkia:
https://github.com/Microsoft/winfile/blob/master/src/wbbitmap.bmp vs. https://github.com/Microsoft/winfile/blob/master/src/bitmap.bmp

Anixx:
So, what should I do to make them black?

Anixx:
@NazmusLabs It is even running under the same OS (Win 8.1)

ttsirkia:
The original branch contais the black version of the images:
https://raw.githubusercontent.com/Microsoft/winfile/original_plus/src/wbbitmap.bmp

craigwi:
this was changed sometime ago (i.e., "before my time").  When I looked at the two bitmap files (wbbitmap.bmp and bitmap.bmp) it appears that the differences are non-trivial.  If someone wants to propose a specific change to revert to the older look, go for it!

Anixx:
I am not sure this should be changed if we are still waiting for someone to fix it.

NazmusLabs:
@Anixx yes, I was actually planning to test out the older bitmap to see how it looks. But trying to figure out the High DPI solutions took my focus off of this. But, now that you reminded me of this, I guess I'll take a shot at doing this now, Insh'Allah, and report back the results, if the application compiles. :)

ttsirkia:
The fix is trivial as the original_plus branch contains the original bitmap with black icons.

Anixx:
Or, alternatively change the color in the new one (I don't think changing one fixed color should be difficult).

NazmusLabs:
@ttsirkia ouch! does it?? Well I should have thought about this before, given that I spent about an hour updating the image pixel by pixel to create a new, sharp, and more pleasing looking version. I'll need to see how my hand made version compares with the one in the Original_plus. I ensured that no existing details while decided to take some creative approach to add some additional details I felt would look better than how it originally looked (judging by the first picture you posted earlier).

[...]

PR #124, opened 2018-04-22
Increased quality of icons used in the app - Address #54

NazmusLabs:
Icons outlines have been made darker, returning it back to mostly the way they used to look in Original_plus version and earlier. This PR responds to multiple people showing preference for darker icon outlines, as can be seen on Issue #54.

The screenshots below demonstrates how the look of the icons will be affected with this commit:

Before:

winfile before with gray file outline

First revision:

winfile after with sharp dark file outline

Second revision (current):
winfile icon sprite updates revised

The second revision was made to address feedback from @craigwi.

Note, however, that the icons are not a complete revert to the old look. It's an evolution of the latest bitmaps used in the master branch to increase sharpness and clarity of the icons. Additionally, the directory tree line color has not been reverted to the old look intentionally.


craigwi:
The changes for folder icons look good to me.  I'm not sure about the document icon changes; many icons I see (e.g., pdf, xls) are not updated (of course) and are of the muted style.  The update made here makes them look out of place.  Thoughts?

Anixx:
Why the tree line color is not changed? I think I would prefer the black one.

Biswa96:
In my opinion, though the icons becomes clear/shiny with these changes, i get some retro feeling with those old icons 😋

NazmusLabs:
@craigwi actually that's a very fair point. The document icon now looks out of place with other native ones. Need to fix that. Thanks!

@Anixx it doesn't look good when you have a black tree line when a lot of folders are open. You get a bunch of parallel lines that looks very distracting. Plus, a grey tree line is the standard in modern operating systems.

NazmusLabs:
@craigwi I've updated the bitmap to make the file icons more consistant. Here's how it looks now.
winfile icon sprite updates revised

@Anixx So, to give you a visual understanding why I chose not to restore the tree line colors to the black used in Original_plus and earlier, I am attatching this screenshot to demonstrate that the current, subtler, gray color is better and less distracting.

winfile original_plus black tree line

craigwi:
Looks good to me.

PR #129, opened 2018-04-23
Retro Branch Applicaton Version is now 9.0.1804.0

NazmusLabs:
Changed the major version number or the application from 10 to 9. Minor version numbers remain unaffected.

Merged latest commits from master branch, as they are apparently known to not have compatibility issues with older releases of Windows.

### DIFF
--- a/src/Winfile.vcxproj
+++ b/src/Winfile.vcxproj
@@ -278,6 +278,9 @@
     <ClCompile Include="winfile.c" />
     <ClCompile Include="wnetcaps.c" />
   </ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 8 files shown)


craigwi:
@NazmusLabs, what type of merge do you want me to perform?  Squash and merge?

craigwi:
After thinking about this, my preference would be for two PR: 

1. merging master with minimal to no changes
2. functional changes in retro

This will keep the history cleaner, I think.  

Thoughts?

NazmusLabs:
@craigwi I'd probably add a third one, though it might fall under functional changes.

3) Undoing commits from master that are incompatible with older versions.

PR #132, opened 2018-04-25
Don't index through reparse points in case they go in an infinite loop.

### LINKED ISSUE #131: Indexer can get caught in an infinite loops of NTFS junctions or symlinks
It is possible to use `mklink` and other tools to create a symlink pointing to a parent folder. If these sorts of directories are traversed naively, it is possible to get caught traversing a never-ending directory hierarchy.

winfile's indexer appears to have this problem. There are a couple of obvious solutions:
- Don't traverse past reparse points. This is easy to implement, but prevents indexing through all junctions and symlinks.
- Detect if a directory has been visited before. This is more work to implement, but it would allow indexing more directory structures.

AustinWise:
Fixes #131 by simply ignoring reparse points when indexing.

### DIFF
--- a/src/wfgoto.cpp
+++ b/src/wfgoto.cpp
@@ -351,6 +351,13 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes, 
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


craigwi:
@AustinWise, I need to revert this change as it doesn't index normal things like my OneDrive folder.

Please look for a more robust solution.

Issue #74, opened 2018-04-10
Show files/folders with the Encrypted attribute in green


FairPlay137:
In Explorer, files and folders with the Encrypted attribute show up as green, as long as you have the "Show encrypted or compressed NTFS files in color" folder view option checked. No such behavior exists here currently, since Windows NT 4.0 was from before NTFS 3.0 (Windows 2000). It'd be nice to see this functionality here. Below are two screenshots showing the folder coloring in Explorer, then in Winfile.
![windows explorer behavior](https://user-images.githubusercontent.com/10647627/38583570-48c0feea-3cc8-11e8-8140-1210c9eb8251.PNG)
![winfile behavior](https://user-images.githubusercontent.com/10647627/38583571-48d56c36-3cc8-11e8-8a30-585abd1a811e.PNG)

Loridawn99:
> ****


>

Loridawn99:
- [ ] 

- [](url)

ZanderBrown:
@Loridawn99 your comments don't seem to be coming through

craigwi:
Fixed in PR #133.

PR #133, opened 2018-04-25
Display encrypted files in green (#74)

malxau:
This is to address issue #74 .  In keeping with the apparent goal of this project, this PR does the minimal thing to address the issue.  Looking at this reveals a lot of housekeeping work though.

1. In the years since Winfile was maintained, a lot more file attributes have been added.  Winfile has been masking in file system attributes with its own internal bits, meaning that it can't represent many of the new file system changes from the last decade without putting its own bits somewhere else.
2. There's a documented bug where it assumes that GetFileAttributes returning a value including 0x8000 isn't a directory (in WFIsDir.)  I don't know the origin of that constant, but today it's FILE_ATTRIBUTE_INTEGRITY_STREAM and it can be on a directory (like compression or encryption) to indicate newly created files should have CRC validation (on ReFS.)
3. This change does not attempt to update the multi-file properties UI to allow encryption state to be changed.  Like compression, this cannot be changed via SetFileAttributes and requires a separate call to EncryptFile/DecryptFile for this task.  If the goal is to also support legacy (pre-2000), this implies a LoadLibrary/GetProcAddress dance.
4. The existing code already pointed out that the colors are hard-coded, which seems bad in light of other work to provide theming, or that the background color or regular file color is arbitrary.


craigwi:
Nice work.  I would like to see two small additions:

1. add an "E" at the end of the global szAttr and change PutAttributes() to use it.
2. add a field to the dialog MULTIPLEATTRIBSDLG to show the Encrypted attribute, but leave the field disabled to indicate it can't be changed.

NazmusLabs:
Quick question: I know that File Explorer can highlight compressed files in blue. Does Winfile do this? If not, we might need to add that eventually.

Also, we would probably need an option to enable or disable color highlighting as File Explorer does.

craigwi:
Actually it is WinFile that shows compressed files as blue; explorer does not show any color (at least on my system) for compressed or encrypted files. Explorer does show a different icon for those types of files.

malxau:
Explorer displayed compressed files in blue for a long time; the icon is a recent thing (Win8?)  Since compression arrived in NT 3.51, I'd go ahead and guess winfile had it first.

As far as the rest, I guess I'm struggling with what the scope of a good PR is.  This project is fairly new (in the team sense) so there's still some things to figure out about how to work effectively together.  I made this PR intentionally minimal because I thought that was the request (from "what makes a good PR": "Can we keep PR focused on one type and one issue at a time?")  But given the comments from three people, there's no dispute that there are a lot of related enhancements to make - the question is, do you want me to submit all related changes in a single PR?  (Think carefully/be careful what you wish for.)  I don't disagree that encryption support would be much better with multi select properties UI, displayed attributes, a persistent option to disable color, or color selection, etc.

Also, I spent a fair amount of time tracking down what happened in WFIsDir.  DOS file attributes query (mov ax, 4300, int 21h) returns 16 bits of attributes or 16 bits of error, in different registers.  16 bit WinFile had an extensible model for network redirectors, which might support long file names but couldn't express them through the DOS interface, so winfile could optionally call a DLL that also returned 2 16 bit values.  Internally though, it encoded this into one 16 bit value with the high order bit indicating whether the remaining 15 bits were errors or attributes; this was feasible because neither error nor attribute numbers could go to 16 bits.  When moving to 32 bit land, all of this code was removed, the call to the helper was replaced with GetFileAttributes (which handles long file names implicitly), and the comment was added asking why the flag check is there.  Point is, the flag check is now wrong and I'm certain is superfluous/historical.

malxau:
Are you then specifically saying that @NazmusLabs request for coloring to be optional not be part of this?

malxau:
Note the latest commit only updates the English dialog template (I don't know the translations for anything else.)  Is there a better way to flag that translation updates are needed?  I did verify that the code functions if the dialog template isn't updated - it'll hit errors trying to update things internally but none of it is visible to the user.

peterooch:
@malxau just add the additional line in every currently available translation, then the translators would translate it.

craigwi:
+1 to @peterooch's comment above.  One more minor item: INVALID_FILE_ATTRIBUTES == ((DWORD)-1).

I'll merge as soon as you have a chance to fix those two things.

PR #136, opened 2018-04-27
add some ignores for mingw build

mattn:


jentur-zabbeJ-8basdy:
> No description provided.

PR #137, opened 2018-04-27
Japanese translation

mattn:
add japanese translation. I looked to remove bJAPAN/bKOREAJAPAN/JAPANBEGIN/JAPANEND. But it is need to keep behavior for users who want to use english on japanese locale.

fe848e3586cb1cb4


mattn:
Please want merge. Some dialogs still have problems for sizes.

image

mattn:
Thanks!

fetus-hina:
Great work.
But this translation does not appear to follow [Microsoft style guide](https://www.microsoft.com/ja-jp/language/StyleGuides).

According to the style guide,

- We should use a space between katakana-words. e.g. "ツール _(space)_ バー" instead of "ツールバー". (Style guide §3.1.7)
- Today, we should use ”－" for English suffix "-er" or "-or", e.g. "ファイル _(space)_ マネージャー" instead of "ファイルマネージャー" or "ファイル _(space)_ マネージャ". ("Katakana prolonged sound mark" of §3.1.14)

If you want to make a nostalgic style (win3.1, win9x and old NT style),

- We should use Half-width katakana instead of Full-width katakana for menu bar items, e.g. "ﾂｰﾙ(space)ﾊﾞｰ" instead of "ツールバー" or "ツール(space)バー".
- In this case, we shouldn't use katakana prolonged sound mark for "-er" or "-or", e.g. "ﾌｧｲﾙ(space)ﾏﾈｰｼﾞｬ" instead of "ファイル マネージャ" or "ファイル マネージャー".

Note: The rule of katakana prolonged sound mark ("ー" for "-er" or "-or") has changed in 2008.

mattn:
@fetus-hina Thanks your advice. I'm not familier to the guide since MS changed style recently after I learned the older style (no spaces, no suffix -er).

Could you please correct words in curent translation? Below is a file listed japanese katakana words in resource files.

https://gist.github.com/4f01bb666a0b1b5af9bf8aa71a869acf

mattn:
@fetus-hina If you are going to make PR, please wait for https://github.com/Microsoft/winfile/pull/138 since this PR change words and dialog sizes.

mattn:
Okay, I'll fix with spaces and suffix -er in my self. Sorry.

clzls:
**I think we may not leave Japanese translations in ``res_en-US.rc``.**

Since localization for many countries and languages are working in process, it's not a good idea to contain Japanese translations for them (for example ``res_zh-CN.rc`` and ``res_pl-PL.rc`` now).

@craigwi 
Translators from other countries can't read Japanese words and just make a copy of ``res_en-US.rc`` for their work. So ``JAPANBEGIN`` and ``JAPANEND`` may leave many rubbish in their works.

craigwi:
There seems to be a different of opinion on whether to leave the JAPANBEGIN/END stuff in or not.  My request is for someone who can figure out what is going on to propose a specific change.  Thanks.

Issue #135, opened 2018-04-26
[Breaking bug] Feature that worked in v10 is broken in master


NazmusLabs:
We have had a major oversight somewhere in one of the commits made since this app went open source. I just recently noticed it and decided to see if the problem exists on v10.0.0 and the Original_plus versions, and it does not.

## The Affected Feature
The bug affects what how Winfile should respond when a user selects a folder in the directory tree that they don't have the permissions to access

### Intended Behavior
This is how it works (and is supposed to work in master) in v10.0.0.0 and original_plus. Selecting a folder from the left directory the without having the permissions to access it displays a message on the right saying that the user doesn't have the correct permission to accress the folder.

### Behavior in Master Code
In the latest build compiled from the master branch, selecting a folder without sufficient permission throws a critical error message that just be closed, after which a message appears on the right staying that an unknown error occurred. No error dialog should pop up and the message on the right should correctly identify the reason for not being able to access the folder rather than saying an unknown error occurred.

## Solution
We have had dozens of commits since v10.0.0.0. We need to first identify which commit has caused this issue to appear.

It's probably best to start by looking at commits dealing with the core functionality of the application rather than things like visual styles.

Another approach is to build the application from an earlier snapshot of master to see if the bug exists and continue doing so until we find where this issue was introduced.

### This is a braking bug; master not fit for release
We cannot generate a release build of Winfile until this bug is fixed. It would be a wrong thing to release the app with such a feature broken. It breaks work flow. For instance, I cannot scroll down my directory tree with the arrow keys without the error popping up, breaking my work flow.

If anyone finds the commit that caused this, please mention it below. Once we identify it, we can start working to revise that commit to fix the issue.

craigwi:
We should definitely get to the bottom of this.  However, I am not able to repro this yet.  In the common case of "C:\System Volume Information", the master branch shows the expected (left picture above). 

What permissions are set for the directory "Delivery Optimizations" in your example?

craigwi:
I checked v10.0, x64 and the bug exists there (i.e., I created it when I added the Wow... stuff).  Thanks @malxau for finding and fixing!

PR #140, opened 2018-04-27
Don't return ERROR_INVALID_FUNCTION on x64 builds for enumerate errors (#135)

malxau:
Looks like WOW redirection on x64 causes LastError to be reset, and the directory enumerate code was expecting ERROR_ACCESS_DENIED not ERROR_INVALID_FUNCTION.  Note this is specific to x64.

~~(I'm not thrilled about how this works when FindFirstFile finds something that the filter wants to remove.  Previously LastError in this case seemed random, now it returns FALSE with LastError of zero.)~~

When filtering FindFirst will always call FindNext, and FindNext is the thing setting the error.  It also had the same WOW bug though.

### DIFF
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -140,7 +140,7 @@ WFFindNext(LPLFNDTA lpFind)
       }
 
 	  Wow64RevertWow64FsRedirection(oldValue);
-
+      lpFind->err = 0;
       return TRUE;
    }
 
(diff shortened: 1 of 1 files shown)


src/lfn.c
```diff
@@ -144,9 +144,9 @@ WFFindNext(LPLFNDTA lpFind)
       return TRUE;
```
craigwi: to be consistent, we should set lpFind->err = 0 just before returning TRUE.

PR #145, opened 2018-04-28
Set application manifest for targeting Windows 8.1 and Windows 10

MouriNaruto:
In Windows 8.1 and Windows 10, the GetVersion and GetVersionEx functions have been deprecated. In Windows 10, the VerifyVersionInfo function has also been deprecated. While you can still call the deprecated functions, if your application does not specifically target Windows 8.1 or Windows 10, you will get Windows 8 version (6.2.0.0). 

In order to target Windows 8.1 or Windows 10, I think update the application manifest in the source file is necessary.

Reference: https://msdn.microsoft.com/en-us/library/windows/desktop/dn481241(v=vs.85).aspx


craigwi:
Seems like a reasonable thing to do, but I worry about how this will work on the earliest support builds of master.  Also, I found https://support.microsoft.com/en-us/help/978637/an-application-runs-in-a-windows-vista-context-instead-of-in-a-windows.

Compatibility gurus, what do you think?

thecatkitty:
Maybe I am not a _guru_, but I used to compile a program running on _all_ Windows versions (I mean, from 1.01 to 10). Windows 2003 and older will just ignore the manifest and run _normally_.

I've extracted KB978637 and it seems that this is a patch for the 64-bit Windows mechanism handling the issue of old, mixed 16/32-bit setup program launchers. It has nothing to do with us, unless we use MS-Setup (1997) or somehow integrate with a very old version of NetMeeting. I think that we're not going to do any of that.

NazmusLabs:
@craigwi we should be perfectly safe with this PR. Microsoft documentation states that adding this manifest has no known compatibility issues with previous versions of Windows.

According to Microsoft:

"Manifesting the .exe for Windows 8.1 or Windows 10 will not have any impact when run on previous operating systems. You can also add this to your .rc file if you already have it defined.
Adding the trustInfo isn’t essential, but it is highly recommended. This will allow your .exe to always get the correct version, whether the operating system is Windows 10, Windows 8.1, or Windows 8."

craigwi:
You folks are great!  Thanks for the confirmation.

PR #146, opened 2018-04-28
fix build on mingw

mattn:
SUBLANG_XXX is defined in winnt.h

And I modified res_ja-JP.rc .


src/res.rc
```diff
@@ -11,6 +11,7 @@
 
 #define APSTUDIO_HIDDEN_SYMBOLS
 #include <windows.h>
+#include <winnt.h>
```
craigwi: Windows.h includes windef.h which includes winnt.h.  Is this not true for mingw?

mattn: mingw resource compiler seems not look winnt.h so SUBLANG_XXX is not defined. I found workaround for this now.

PR #147, opened 2018-04-28
quote filename for edit

mattn:
Some text editor treat the argument which contains space as multiple arguments. ex: gvim.exe


mattn:
Thank you

PR #148, opened 2018-04-28
Add Hebrew translation

peterooch:
Some of the messages taken from the WfW3.11 Hebrew version
Switched all references of directories to folders (using modern terminology)
`DIALOG` -> `DIALOGEX` to enable Right-to-left layout for dialogs


src/Winfile.vcxproj.filters
```diff
@@ -102,6 +102,15 @@
     </None>
     <None Include="lang\res_pl-PL.rc">
       <Filter>lang</Filter>
+    <None Include="lang\suggest_he-IL.db">
```
craigwi: The <None></None> got mismatched.  The stray </None> from below needs to be moved up here.

Issue #99, opened 2018-04-15
winfile does not support UNC paths


n3rdopolis:
Hi.

Attempting to browse to a UNC path with winfile, (with file>goto directory) it seems that winfile doesn't know how to handle it. 

Thanks

craigwi:
Yes, that is the current behavior.

rinseaid:
I would like to see this feature implemented.

Serentty:
With the WSL, this is more important than it used to be, since Linux drives don't get letters by default, but rather show up as UNC paths.

Serentty:
In case anyone else ran into the issue of accessing WSL files, you can use the ``subst`` command to get a drive letter for a UNC path, like this:

```
subst Z: \\wsl$\Arch
```

schinagl:
> Yes, that is the current behavior.

The odd thing is: winfile is super close to handling UNC path. 

I have a private version which perfectly navigates on a large UNC drive without problems. 

There are little changes needed
- FindItemFromPath needs some extra care. Ok it took a Sunday morning😉
- Two lines of code in FillTreeListBox
- IsRemoteDrive needs an if statement to treat UNC
- and a hotkey which allows to paste UNC path into winfile. 
- the 1024 path extension from #340 (more for convenience, because UNC path tend to be huge)

Well with this one can navigate/copy/rename/.... most tasks. 

Currently it is a proof of concept, but once it is tested, I will at least provide draft PR, so that others could play.... if there are others, which still use this😆

PR #152, opened 2018-04-29
Rename HTML documentation files using the 8.3 filename convention

thecatkitty:
@NazmusLabs has converted the CHM help from Windows 98 into discrete HTML pages.

I just made it open in Windows 3.x.

Windows File Manager HTML help in Internet Explorer 5 on Windows for Workgroups 3.11


NazmusLabs:
Thanks! Very much appreciate the changes! I'll close the original one, now that this is online! 😊

craigwi:
Question: what are the references to "Unknown" in the files?  One appears to be an author statement.

NazmusLabs:
I used a tool to convert the Metadata included with the CHM file to HTML, and as the CHM files had some Metadata missing, the tool just put an "Unknown" there instead.

thecatkitty:
I have more than one question:

1. Are we going to edit these docs in the near future?

2. Are we going to replace _Unknown_ with something or to drop it whatsoever?

3. If we're going to expand the docs, we'll need to consider the whole format of them. What do we want them to be like?

4. Are we going to translate the docs in the future, too?

5. Aaand... A question which is recurring among other contributors and disputants: Do we have any broad vision of what we want to achieve? Are we going to deliver just a nice tool for the passionate, as a secondary file manager (just like a certain tool with the word _Total_ in its name), or rather a full-fledged (but program-manager-less) Windows Explorer/File Explorer replacement capable of getting things done on Windows PE and Server Core?

craigwi:
I remain curious about 2 ("unknown") and 5 (the "big" question).   I see WinFile as a "nice tool for the passionate".  I use it as my primary file manager and so don't identify with the idea that it is a "secondary file manager".  

However, my primary question is what features would a "full-fledged" explorer have that WinFile does not have?  When I find things missing, I just add them.

ad43:
>  I use it as my primary file manager and so don't identify with the idea that it is a "secondary file manager".

Well, what does functions does Win10 Explorer do that File Manager doesn't? Explorer removed the option to open 2 or more windows onto the file system and save the configuration. So perhaps File Manager should be able to save more than one window config (I mean MDI config) to make File Task management easier. Explorer adds ghost 'Library' and 'Network' etc entries to the filing system tree which are navigational aids and not actually files or folders. In my opinion they confuse things immensely and should reside in a Program Manager view or a My Stuff or Desktop view or a properly configurable and stable Start Menu, not in a File Manager.

I do enjoy the right-click Actions menu though.

Chris

craigwi:
I'm aware of what Win10 explorer does, but I don't understand what @thecatkitty means by "a full-fledged (but program-manager-less) Windows Explorer/File Explorer replacement capable of getting things done on Windows PE and Server Core?"

thecatkitty:
The main reason I've written such a thing was that I've been wondering if we want all the File Manager functionality to run properly in environments lacking the Explorer completely, like Server Core. If that is the case, there could be situations when re-inventing things would be needed.

clzls:
@craigwi 
>I'm aware of what Win10 explorer does, but I don't understand what @thecatkitty means by "a full-fledged (but program-manager-less) Windows Explorer/File Explorer replacement capable of getting things done on Windows PE and Server Core?"

First, I think `Windows PE` here includes `Windows Recovery Environment` as well. Windows RE has no task bars, Start Menu, even `explorer.exe` process and so on. We certainly need a handy tool to manage it.
Second, I think these are some topics:
- `Plug & Play` support;
- `Properties` viewer (Rely on Explorer now);
- Access security managements (Also rely on Explorer now; I'm not sure whether it is this name in English; maybe authority control or similar things that controls how a certain user/group can access certain file(s)...);
- UNC path (#99);
- Help system;
- Further thoughts (Making it a __total__ replacement of `Explorer.exe`, but maybe not necessary, I think):
  - Power control: shutdown, reboot, ...
  - `Format Drive` and total disk operation support (Rely on Explorer now);
  - Task bars, Start Menu, ... or similar solutions;
  - Desktop icons, ...

I think that the lack of `Plug & Play`(not `autorun`, but the ability to detect new drive on the system) support for removable devices is the first topic of them all. You must restart winfile in order to access new devices under current version; that's quite unnatural.

PR #156, opened 2018-04-30
Solving warning c6400

DBJDBJ:
to solve (few ot these):

d:\github\winfile\src\wfutil.c(859): warning C6400: Using 'lstrcmpiW' to perform a case-insensitive compare to constant string 'NTFS'.  Yields unexpected results in non-English locales.

I have created winfile_internals h and c

idea is to keep (re usable) solutions in there ..

as for C6400, I would not want to apply it to the code base because I am not sure when to use 

ordinal string comparison 

or when to use

locale sensitive string comparison

there are not that many places in the code where this needs to be applied so if someone points me where to use which I can do that too ..

code is properly commented.

### DIFF
--- a/src/BagOValues.h
+++ b/src/BagOValues.h
@@ -89,7 +89,8 @@ namespace winfile {
 		using lock_unlock = internal::lock_unlock;
 	public:
 		typedef T value_type;
-		using storage_type = std::multimap< std::wstring, value_type >;
+		typedef std::wstring key_type;
+		using storage_type = std::multimap< key_type, value_type >;
 		using value_vector = std::vector< value_type >;
 	private:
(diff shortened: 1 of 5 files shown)


DBJDBJ:
ps After applying this fix one would want to test it on several machine with different non English locales.

DBJDBJ:
I do not know why it appears I have changed anything on this branch of mine beside adding two new files 
winfile_internals.h and winfile_internals.c ? I have branched form microsoft/winfile master.

DBJDBJ:
to narrow the solution scope I need to close this PR , remove unwanted changes and then re issue a new PR about the C6400, solution.

PR #157, opened 2018-05-01
add missing japanese translations

mattn:
Also convert file format to UNIX. You can see diff without changes of whitespaces at https://github.com/Microsoft/winfile/commit/9697249f0cdc290c30e3fd48d5af5eb8df8c7449?w=1


craigwi:
WinFile uses DOS line endings (\r\n).  Also, as noted in #88, I do not want to see formatting mixed with functional/translation changes.

mattn:
Current files in src/lang use UNIX line endings. And only ja_JP use DOS line endings. Should I keep it so?

craigwi:
You are right; what is in the repo is UNIX line endings.  This is probably because I use core.autocrlf == true.

Then my request is to make two PR, one to change the line endings and one to change the content.  Thanks.

craigwi:
Hi @mattn.  I fixed some of the inconsistencies in line endings, but I would not go to the trouble of converting to \n in the repo.  It appears I have been inconsistent in this manner.  

Can you fix your PR to reflect just the changes to the content?  Thanks.

PR #159, opened 2018-05-04
Removing About item from popup.

artursouza:
This menu item seems useless in the popup. It is already available in the top menu items.

### DIFF
--- a/src/lang/res_en-US.rc
+++ b/src/lang/res_en-US.rc
@@ -144,7 +144,6 @@ BEGIN
 	MENUITEM    "Start Cmd She&ll...", IDM_STARTCMDSHELL
 	MENUITEM    "Start Po&werShell...", IDM_STARTPOWERSHELL
 	MENUITEM    "&Goto Directory...", IDM_GOTODIR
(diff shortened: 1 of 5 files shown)


clzls:
@craigwi Should this PR be accepted?

artursouza:
I just found weird to have a pop-up menu item with "About". I usually see
this in the top menu only - why my argument that removing this item should
make the UI cleaner without removing access to the About dialog ("it is
already available in the top menu items").

NazmusLabs:
I think the "right-click" faux context menu should only have menu items that deals directly with file management. So things like copy, move, properties, format makes sense. That's the purpose of a *context* menu. It shows items relevant to the context where you right click.

About is an app functionality that isn't contextual to file management. So it should be removed.

In fact, we should take a look at the context menu and remove anything else that doesn't deal with file management and add anything that isn't there.

For instance, the sort option SHOULD be on the context menu, as well as view options.

craigwi:
Reasonable.  Thanks @NazmusLabs for chiming in.

artursouza:
Thanks everyone for the good discussion. I will look for more interesting
changes to follow.

PR #162, opened 2018-05-06
Support 32 bit MinGW Build

### LINKED ISSUE #161: Support 32 bit MinGW Build
Currently only the 64 bit build of WinFile works with MinGW. A 32 bit build fails for the following reasons:

  1. The code uses Windows Vista functions but does not set the WINVER preprocessor definition.
  2. The code has casts between incompatible function pointer types for callback functions, which silences compiler errors and can cause crashes at runtime due to stack corruption (this works fine in MSVC because the project sets the default calling convention to `__stdcall` so there is no mismatch).

A PR will follow shortly.

MartinPayne:
Set WINVER preprocessor definition and corrected declarations of callback functions.

Closes #161

### DIFF
--- a/src/GNUmakefile
+++ b/src/GNUmakefile
@@ -34,7 +34,7 @@ SRCS = \
 
 OBJS = $(subst .c,.o,$(SRCS)) wfgoto.o res.o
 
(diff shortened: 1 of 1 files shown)


craigwi:
@MartinPayne, really nice work!

Issue #158, opened 2018-05-03
'for' loop initial declarations in wfcomman.c:538:5


clzls:
'for' loop initial declarations are only allowed in C99 mode. So maybe we should use option -std=c99 or -std=gnu99 in Makefile?
In my opinion it's more like a "modern" C++ style of coding, not a style decades ago. (for example before 90s)

PS. I'm using `mingw32-gcc (GCC) 4.7.2`, a rather old version(I thought it may work for these codes over 20 years old...). After bumping into "internal error" in VS2017, I tried MinGW but I still bumped into hundreds of lines of errors and warnings. Maybe it's all my fault...

DBJDBJ:
Microsoft (R) C/C++ Optimizing Compiler Version 19.13.26132 for x86
for C code, compiler "default option" is C99
There is no /std:c99 option
AFAIK. There is no way to dictate which exact C option one wants to use.
This behavior is one of "compiler default" options. Of which AFAIK there is no official list ...

DBJDBJ:
Also there is no __STDC_VERSION__ predefined macro.
So using cl.exe one can not check grammatically which C version is in use.
Like [here ](https://wandbox.org/permlink/FNJ4H31F0p524MJW) for example.

clzls:
@DBJDBJ Or maybe we can just review all the code to satisfy C90 and make VS2010 happy(#165 )?

DBJDBJ:
@clzls AFAIK it is Visual Studio "the latest"  ... I have unleashed code analyzer on winfile and that has revealed itself as a very good method. But this is going to be one very long task. I assume the question is how to manage it. This is all for x86 builds.

x64? Just switching to that mode unleashes ocean of warnings and few errors, too. The process of removing them (for me at least) will be very close to rewrite ...

clzls:
As Craig said in #165 , let's stick with modern compilers and forget my `mingw32-gcc (GCC) 4.7.2` now. So I will close this issue.
For x64 ... maybe we can put it aside and focus on core codes ...

PR #165, opened 2018-05-08
make VS2010 happy

ssrlive:
make VS2010 happy

### DIFF
--- a/src/BagOValues.h
+++ b/src/BagOValues.h
@@ -61,7 +61,7 @@ public:
 	{
 		wstring lowered;
 		lowered.resize(query.size());
-		transform(std::cbegin(query), std::cend(query), std::begin(lowered), ::tolower);
+		transform(std::begin(query), std::end(query), std::begin(lowered), ::tolower);
 
 		vector<TValue> results;
 		TValue val = TValue();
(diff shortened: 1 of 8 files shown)


clzls:
It may be the same idea of #158 . Either we remove all 'for' loop initial declarations, etc. to fit ISO C90 or lower version, or we choose to adopt C99 standard and not to satisfy VS2010.

ssrlive:
I think my commits fit ISO C90.

clzls:
@ssrlive , I mean that we should decide whether to fit C90 or send it to the trash and adopt C99.

ssrlive:
I think it's easy to make the decision. because the code base will not change many and supporting more compilers is a good idea.

craigwi:
This PR is also similar to #59.  The current thinking is that the master branch is compiled using one of two toolsets: VS2017 and mingw.

Right now, it would be easy to make VS2010 work, but the question is whether going forward we support VS2010 indefinitely.  

I agree with a comment @leeter made in #9 and elsewhere: let's stick with modern compilers to get the best code generation, optimization and security features.

Other comments?

clzls:
I second the motion. We should send C90, MinGW 3.x, VS2010s into the trash.
(But I cannot compile it in my VS2017 Community. I bumped into "internal errors" in `BagOValues.h`)

ssrlive:
I have finished the work for Windows XP and VS2010. https://github.com/ssrlive/winfile/tree/winxp 
Bye bye.

clzls:
@ssrlive , WinXP support is in `retro` brunch. Thank you anyway.
Maybe you should read #88 before you start working. See you next time.

PR #166, opened 2018-05-10
Right-to-Left UI

peterooch:
Enables a more streamlined UI direction, added with dialogs being also being in the same direction (in Hebrew translation for now).

Added entry in preferences dialog to enable/disable this feature.
With mirroring disabled it will look like the original win3.11 layout (as shown in the images below)

Issues/Opinions needed:

1.  As seen in the first image, the toolbar icons are flipped, and the only solution I thought of is to provide a secondary bitmap with flipped images to be used if mirroring is enabled.
2.  To add the option in non-RTL locales to enable the UI mirroring? the code is already there, only the dialog resources need to be updated.

With UI mirroring enabled
win10_winfile_rtlook

With UI mirroring disabled
win10_winfile_classylook

Win3.11 Winfile ui
win3 11_winfileheb


craigwi:
I reshaped the PR per my comments in the commit.  Note that the winfile.ini setting changed in name and meaning, but the end effects are the same!   @peterooch, great work!

peterooch:
Just a note, enabling it for arabic locale while not having an arabic translation will mirror the UI (even if in english) unless the language setting in winfile is set diffrently.

PR #169, opened 2018-05-11
Add German translation (mostly from original German version)

BaldwinTechnology:
This is the german version of winfile


BaldwinTechnology:
I see I forgot to set the width of the controls.

craigwi:
The about menu item was removed from the context menu (what is named "Dummy Popup"); can you remove that from your version?  

Otherwise, look fine to me.

craigwi:
Oh and please do not use tabs; 4 spaces == tab (e.g., wfloc.c).

BaldwinTechnology:
I removed the entry and will add a new pull request as soon as I have corrected the Dialog width.

@craigwi: Where did I used tabs? I checked my files and there are only spaces.

src/wfloc.c
```diff
@@ -20,6 +20,7 @@ LPCTSTR szLCIDs[] = {
     TEXT("he-IL"),    // Hebrew, Israel
     TEXT("ja-JP"),    // Japanese, Japan
     TEXT("pl-PL"),    // Polish, Poland
+	TEXT("de-DE"),    // German, Germany
```
mattn: please fix indentatino

BaldwinTechnology: Oh, I see.

craigwi:
Fixed in #176.

PR #175, opened 2018-05-17
Fix GNUmakefile

mattn:


craigwi:
Thanks @mattn; sorry to have sprung this need on you...

PR #177, opened 2018-05-25
Disable "Edit" menu item for directories.

artursouza:
The "Edit" menu item does behave the same way as the "Open" menu item for directories. On the other hand, I was not expecting this menu item to be enabled. The "Print" option is disabled for directories, for example. This change is to add make the "Edit" menu option behave the same way as the "Print" option. This change also handles to context menu.


src/lang/res_ja-JP.rc
```diff
@@ -144,7 +144,6 @@ BEGIN
 	MENUITEM    "Cmd シェルを起動(&L)...", IDM_STARTCMDSHELL
 	MENUITEM    "PowerShell を起動(&W)...", IDM_STARTPOWERSHELL
 	MENUITEM    "別のディレクトリーへ移動(&G)...", IDM_GOTODIR
-	MENUITEM	"ファイル マネージャーについて(&A)",			IDM_ABOUT
```
artursouza: Fo some reason, this change was reverted so this PR re-applies this change.

Issue #194, opened 2018-08-05
Stack overflow while building goto cache when encountering self-referencing link with empty cFileName


bytes-and-bits:
I have access to several Win7 machines which have security software that creates a link with an empty name that points to itself - probably some kind of attempt to foil directory-crawling malware.  When running the latest release of Winfile I get a stack overflow when it attempts to traverse the link while building the goto cache.  Windows Explorer completely ignores the link so I don't see it there, while Winfile will show the link with a blank name, and it lets me traverse it down multiple levels.  Winfile should at least ignore those links when building the goto cache, if not ignore them globally.

![dir_structure](https://user-images.githubusercontent.com/8423814/43682824-1de72362-9834-11e8-9358-fedff371daa0.png)

![empty_cfilename](https://user-images.githubusercontent.com/8423814/43682865-032fc0aa-9835-11e8-9f19-842d4804b71c.png)

![stack_overflow](https://user-images.githubusercontent.com/8423814/43682866-05b9e9f4-9835-11e8-832d-d38094e75392.png)

bytes-and-bits:
Fix verified in the 1901 release.

PR #195, opened 2018-08-05
Ignore directories with blank names when building goto cache

bytes-and-bits:
Ignoring directories with blank names while building the goto cache fixes the crash reported in https://github.com/Microsoft/winfile/issues/194


craigwi:
Fixed in 9014cf47d12132c49a8c39910025ed325b53dd14 and attributed to @bytes-and-bits.

PR #204, opened 2018-10-15
Update wfloc.c to add ko-KR

jaeheonshim:


craigwi:
To enable a language, you need to provide three language-specific files in addition to the change in wfloc.c and other changes:
1. res_ko-KR.rc
2. suggest_ko-KR.db
3. winfile_ko-KR.dlg

You are probably best off starting with US English (en-US) and translating from there.

PR #206, opened 2018-10-25
Getting ExeView to run in x64

misirlou-tg:
Fixing calls to SendMessage() that cast the last parameter to LONG, changing to LPARAM.

The one that was causing ExeView to crash at startup was in init.cpp, SendMessage with LB_SETTABSTOPS.


craigwi:
Good catch.  Thanks.

PR #207, opened 2018-10-27
Create READMES.md

Santosh-Sah:


craigwi:
This seems like a feature request.  Why are you suggesting adding a file with two lines?

Santosh-Sah:
Yeah, initially I thought changing file menu language should be good then I realize its not a bug rather be a feature. Kindly reject my pull request, I will do later on.

PR #209, opened 2019-01-05
Japanese Version for the README.md file and Some Changes in src/lang/res_ja-JP.rc

Takym:
I translated the README.md to Japanese and I changed some parts of src/lang/res_ja-JP.rc. I was wondering if you could check my README.ja.md file and my changes in src/lang/res_ja-JP.rc.


clzls:
About the translation changes:
Though I can't read Japanese, I would like to remind you to follow Microsoft Style Guide and discussions made by other Japanese translators (for example #137). Be careful about changing Katakana to Kanji (or any other changes) unless most of Japanese translators agree with the change.

PR #211, opened 2019-01-17
Use HTML Help instead of Windows Help

### LINKED ISSUE #184: Help menu opens browser instead of included HTML help file
This command needs to be updated to accommodate the change from HLP to CHM.

thecatkitty:
This change makes WinFile open CHM file instead of HLP. This resolves #184. In addition, it removes _Help on Help_ command from the menu (there's no such HH functionality).

The code assumes that `winfile.chm` is placed in `x86` directory, which is true for ZIP (not for 1901, apparently...) and Chocolatey releases.


Amaroq-Clearwater:
@thecatkitty Thank you for fixing this. I would like to merge this pull request into my own fork, which aims to add transparent UTF-8 support.

NazmusLabs:
image

I highly recommend this. As per Microsoft's own Windows app guidelines, apps using the old help technology should migrate to HTML help as a direct migration path from the now unsupported help system.

src/winfile.h
```diff
@@ -1342,8 +1342,9 @@ Extern LARGE_INTEGER qTotalSpace;
 
 Extern HWND hwndStatus        EQ( NULL );
 
-Extern TCHAR szWinfileHelp[]  EQ( TEXT("WINFILE.HLP") );
-Extern WCHAR wszWinfileHelp[] EQ( L"WINFILE.HLP" );
+Extern TCHAR szWinfileHelp[]     EQ( TEXT("..\\X86\\WINFILE.HLP") );
```
malxau, line 1345: Why are these path changes needed/related?

PR #215, opened 2019-02-01
Added store link to ReadMe

kolappannathan:


README.md
```diff
@@ -16,6 +16,11 @@ I will not be changing the original_plus sources nor creating other branches for
 You are welcome do that on your own.
 
 ## Download The App
+
+If you are running Windows 10, you can download the app from the store
+
+<a href="https://www.microsoft.com/store/apps/9p7vbbbc49rb?ocid=badge"><img src="https://assets.windowsphone.com/85864462-9c82-451e-9355-a3d5f874397a/English_get-it-from-MS_InvariantCulture_Default.png" alt="MS Store Badge" width="150px"></a>
```
HowardWolosky: I see that you created this via the [generator](https://developer.microsoft.com/en-us/store/badges), but I think you'd probably want to use standard markup for this:

```suggestion
[![Get it from Microsoft](https://assets.windowsphone.com/85864462-9c82-451e-9355-a3d5f874397a/English_get-it-from-MS_InvariantCulture_Default.png "Get it from Microsoft")](https://www.microsoft.com/store/apps/9p7vbbbc49rb?ocid=badge)
```

> Although, you can't specify size with standard/GitHub-flavored Markdown [as far as I can tell](https://stackoverflow.com/questions/14675913/changing-image-size-in-markdown), so maybe this format should stay as-is.

kolappannathan: > Although, you can't specify size with standard/GitHub-flavored Markdown as far as I can tell, so maybe this format should stay as-is.

Well my previous format actually formats the size. The applied suggestion removes the size limit and makes the images go full width.

kolappannathan: The other answers in the link you provided point out that using HTML in place of markdown allows you to resize the image.

kolappannathan: Should the image be shown to full width which is extremely large so that the doc sticks with the standard markup?

PR #220, opened 2019-03-19
WINFILE.INI's location need to be flexible to be more useful

zhw2101024:
WinPE don't have APPDATA environment variable, thus WinFile is unable to write and read config file. Seems USERPROFILE environment variable works, not sure if older systems have this environment variable. So let it be the second choice now.
Appreciate for any advice or discussion


zhw2101024:
a portable version automatically detect and use ini file in the same folder may be better

PR #223, opened 2019-05-06
Image file updates (improved high-res icon, prototype symbol, etc)

Amaroq-Clearwater:
Renamed LICENSE to LICENSE.TXT, and gave the high-resolution icon a spit-shine and polish so that it's less visually noisy. In other words, I fixed the dithering.


Amaroq-Clearwater:
Edit: The BMP version of the prototype symbol I made has lost its alpha channel for some reason. Thankfully, I've already provided the PNG version, whose alpha channel is intact.

ZanderBrown:
You seem to have removed the grid texture thing, which IMO is a feature not a bug

Amaroq-Clearwater:
It was dithering. Dithering is noisy and distracting. Simply put, it is unpleasant to look at.

ZanderBrown:
But that's part of the icon design for winfile, the design choices of 1990 might be regrettable but that's part of winfiles charm

craigwi:
Winfile256.png is not used anywhere; I created it while working on the Windows Store version.

Except for the new addon image, I don't see any visual differences in the other images (which appear to just be compressed more).  Were there changes in those?

Changing the license file is unnecessary.  Once you get used to the feature of WinFile where F12 loads the file in Notepad (or other editor), you won't worry about extensions much again...

Amaroq-Clearwater:
The other images were re-encoded using a handy utility called PngGauntlet, you can look it up. I was able to make the image files smaller through entirely lossless means. I also plan to replace the various BMP files with PNG files.

@ZanderBrown Regarding the dithering on the icons in general, Windows 3.1 may support 256 colors, but not all computers at the time actually had that capability. They had to work within the limited (and quite ugly) color palettes of EGA graphics for a lot of things.

Amaroq-Clearwater:
Since there doesn't seem to be any interest, I will go ahead and close my pull request.

PR #250, opened 2020-01-31
Change Encoding of src/origrc/winfile.ru-RU.rc to UTF8

clzls:
Change encoding of src/origrc/winfile.ru-RU.rc to UTF8.


clzls:
(There are some commits "ahead of microsoft:master" but they have no changes to microsoft:master)

PR #254, opened 2020-04-22
Modify GNU Makefile to use standard variables

stefan-misik:
GNU Makefiles often use standardized variable names to e.g. allow to
override default values. This allows us e.g. to issue following command
to cross-compile the project using 64-bit Cygwin without modifying the
makefile.

   make -f GNUmakefile \
     CC=i686-w64-mingw32-gcc \
     CXX=i686-w64-mingw32-g++ \
     WINDRES=i686-w64-mingw32-windres \
     LDFLAGS+=-static

For more info see: https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html and https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html#Catalogue-of-Rules .


craigwi:
@mattn, any concerns about this?  Seems pretty simple.

PR #259, opened 2020-08-16
Add appExecutionAlias so 'winfile.exe' appears on the PATH.

thetorpedodog:
Hello! I am new to Windows development so please let me know if there is a better way to accomplish what I'm trying to do here. I am banging rocks together based on UWP docs.

It seems weird that the executable name has to be specified in the extension, but [that's what it looks like](https://stackoverflow.com/questions/55426605/how-to-properly-set-appexecutionalias-so-the-program-could-be-launched-from-comm).

Thank you!


malxau-msft:
I resolved conflicts in this PR and merged it.  Unfortunately Github doesn't see that, but the commit should be credited to Paul.  Thanks!

Issue #260, opened 2020-09-04
Right Mouse Click Context Menu issues


MooseValley:
Pretty certain this is a know issue ... but it wasn't listed in the issues, so here goes ....
The Right-Click context menu is faulty / incomplete:
1. If I select a bunch of files and right mouse click, only the right mouse clicked file remains selected.
2. If I select one or more files and right mouse click, there are no options to "Open With ...", "Create Compressed ZIP", etc.
3. Also, I use 7z and this isn't showing on the context menu, but this is something they need to deal with / support - this is not a FileMan problem ...
 
Anyway, appreciate your work on bringing this great old tool back to life !!

File Manager was the software I missed most when I moved from Windows 3.x to Windows 95.  Having 2 or 3 drives open at a time, arranging them how I liked, etc without the clunkiness of having multiple copies of File Man running (multiple main windows) was sorely missed.  Glad to see this great old software making a comeback.

Moose
Moose's Software Valley, established July 1996.
https://rebrand.ly/MoosesSoftware

craigwi:
All good points.  I noticed 1. before and consider it a bug.  The other points would be a feature add.  I like the idea and welcome a PR for it.

MamiyaOtaru:
without some serious reengineering I think 1 is intractible.  I looked into it to see if I could right click on a dir in the tree to open it in a new window without also selecting it.  wfdlgs ActivateCommonContextMenu handles the right clicks.  It works by selecting the item under the cursor, then bringing up the right click menu.  It doesn't pass along any info about the item that was clicked.  When Winfile gets a command back from the right click menu (open in new MDI window, open cmd window here, etc), how does it know what to act on?  It carries out the chosen action on the (singular) item that was selected in ActivateCommonContextMenu.

If you click on the tree (lines 892,893) LB_SETCURSEL highlights the item, LBN_SELCHANGE then informs other widgets (like the directory window).  Omitting either of those means the right click action is carried out on whatever had been selected before.  To carry out the item ont he action actually clicked, the item has to be selected and it is singular.  

To fix: would need a way to store what item had been clicked on without full on selecting it.  For right clicking multiple items, would need to not select the singular item under the cursor, and know to act on the selected items *if and only if* the item right clicked on was one of them, otherwise should act on the item that was right clicked (again, ideally without selecting it).  

It would add a bunch of complexity that just wasn't there before and I balk a bit at what it would take

MamiyaOtaru:
[...]

PR #271, opened 2021-06-05
Adding explorer right-click context menu

misirlou-tg:
This is a provisional pull request to show how explorer integration could work, this is mentioned in #260

When right clicking on a file in the right pane Explorer's context menu is displayed instead of the WinFile version


misirlou-tg:
I pushed changes where the explorer menu shows up as a popup from the WinFile context menu. For example:
image

PR #278, opened 2021-11-09
Fix typos

bitigchi:


schinagl:
I am not a Turkish native speaker, but by simply looking at the changes it is really typos that are fixed
So this could be merged.. ?
What do you think @craigwi ?

bitigchi:
Superseded by #346.

PR #279, opened 2021-11-09
Fix casing

bitigchi:


schinagl:
I am not a Turkish native speaker, but by simply looking at the changes it is only a case change and one more letter at the end.
So I guess this can be merged? 
What Do you think @craigwi ?

bitigchi:
Superseded by #346.

PR #284, opened 2022-01-23
Double clicking a relative or absolute symbolic link or a junction in…

schinagl:
### General
Double clicking a relative or absolute symbolic link or a junction in the right directory pane resulted in navigating to bogus directory locations.
Now navigation in the right pane works properly and reparse points are resolved properly.

### How to test
You might test this by running the below .bat file
[WinFileReparseTest.zip](https://github.com/microsoft/winfile/files/7921422/WinFileReparseTest.zip)
and by downloading ln.exe from https://schinagl.priv.at/nt/ln/ln64.zip and placing it in the same directory as the above .bat file

The bat file creates a simple structure with relative and absolute symbolic links and a junction, so that you can click around with winfile in the right pane. You can also create the dir structure with mklink.exe if you would like to avoid the above ln.exe

A little sidetrack ... check https://schinagl.priv.at/nt/ln/ln.html ;-) it does much more reparse point magic. Sry for the ad ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ In summary v10.0 has the following changes/new features compared to original_plu
 
 1. OLE drag/drop support
 2. control characters (e.g., ctrl+C) map to current short cut (e.g., `ctrl+c` -> `copy`)
-instead of changing drives
+instead of changing drives. Change drives now via CTRL + ALT + letter
 3. cut (`ctrl+X`) followed by paste (`ctrl+V`) translates into a file move as one would expect
 4. left and right arrows in the tree view expand and collapse folders like in the Explorer
 5. added context menus in both panes
--- a/src/wfdir.c
+++ b/src/wfdir.c
@@ -2821,7 +2821,7 @@ UsedAltname:
 
          if (hwndDir) {
 
-            // reparse point; szFile is full path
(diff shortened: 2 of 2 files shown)


malxau (approved):
I tried this out with a handful of cases, and it works for navigating through both the right pane and the left pane.

One minor comment: there's a comment above this block saying "reparse point; szFile is full path".  That's not what I'm seeing in the debugger, and this change is assuming it's not a full path.  Should that comment be updated? Is there a case where it is a full path?

schinagl:
You are right. The comment is wrong, thus I changed it.
Which leaves the question why the code provides a StripPath() a few lines later if it is szFileName == filename only.
But to be honest, I do not dare to change this, since DirGetSelection is a rather low level function and is called from many places, where we do not know the constraints of the invocation.

One nice bonus of this fix now is, that when you have my LinkShellExtension installed, you can open the properties dialog of a reparse point, and chose the additional Link Properties, and see where the reparse point refers to :-)

You are a former NTFS maintainer? :-) Oh my god, I wish, I could ask all the questions, which came up during LinkShellExtension and ln.exe development during the last 20 years. If you want to answer send me an email to my email adress mentioned in on ln.html, if not I can understand this.

PR #287, opened 2022-01-26
TypeAhead is buggy in rare situations

### LINKED ISSUE #286: TypeAhead is buggy in rare situations

schin.jpg
sec1.bla
sec2.jpg
sh.jpg
sim.txt
ssh.ext

If you start typing in the right dir pane with 's' 's' 'h' the selection lands on sh.jpg. This means if the first two characters are the same, the second is swallowed, and it matches sh.jpg. This is a bit od, because if you rely on typing you end up on the wrong filename.

Explorer.exe behaves differently: ... it simply works :-)

Anyhow I already found the place where things go wrong in TypeAhead() in wfutil.c, but haven't found a proper solution, since this hydraulics is really tricky. Will investigate on

schinagl:
This solves #286 
You can try this out by executing this little .bat file, which creates a bunch of almost empty files
[GenerateFiles.zip](https://github.com/microsoft/winfile/files/7945562/GenerateFiles.zip)

schin.jpg
sec1.bla
sec2.jpg
sh.jpg
sim.txt
ssh.ext

If you start typing in e.g. the right dir pane with 's' 's' 'h' the selection lands on sh.jpg instead of ssh.ext
This means if the first two characters are the same, the second is swallowed, and it matches sh.jpg.
This is a bit od, because if you rely on typing you end up on the wrong filename.

This PR fixes #286 and gives Winfile.exe the exact same behavior as explorer.exe has.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ In summary v10.0 has the following changes/new features compared to original_plu
 
 1. OLE drag/drop support
 2. control characters (e.g., ctrl+C) map to current short cut (e.g., `ctrl+c` -> `copy`)
-instead of changing drives
+instead of changing drives. Change drives now via CTRL + ALT + letter
 3. cut (`ctrl+X`) followed by paste (`ctrl+V`) translates into a file move as one would expect
 4. left and right arrows in the tree view expand and collapse folders like in the Explorer
 5. added context menus in both panes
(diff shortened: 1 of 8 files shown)


craigwi:
When I wrote the original feature for this, the intention was that typing the same character again and again would move to the next item starting with that letter.  I like that behavior and your change removes that.

That is, to handle both the current behavior and solve the problem you have, I think you will need to pass in more context or return enough info for the caller to do the matching in a more sophisticated way.

PR #288, opened 2022-01-27
ALT + Enter does not work on root dirs

schinagl:
ALT + ENTER does not show the properties dialog for root directories
But one can achieve this anyhow via right mouse on the root directory in the tree control by selecting 'Properties' from the context menu.

This change enables ALT + Enter on the root drive in the tree control aka left pane.

Maybe there is a good reason for the existing behavior, but I don't see the point. Maybe it did not work in older Windows versions, but from my naive point of view everything works nicely.

@craigwi: What is your opinion?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ In summary v10.0 has the following changes/new features compared to original_plu
 
 1. OLE drag/drop support
 2. control characters (e.g., ctrl+C) map to current short cut (e.g., `ctrl+c` -> `copy`)
-instead of changing drives
+instead of changing drives. Change drives now via CTRL + ALT + letter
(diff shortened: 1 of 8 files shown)


malxau:
Please rebase your master branch from the upstream one.  The PR looks for the newest common ancestor, then includes all of the merges and unrelated changes after that point.  Rebase will apply your changes on top of the upstream branch, so there's a newer common ancestor, and the only changes in the PR are the ones that differ from the upstream branch.

schinagl:
Sry. It was my git unability, because I just used fetch and merge from the guthub website
Learned now, that one has to do it via rebase.
There are now two new pullrequests

PR #289, opened 2022-01-28
Inherit ACLs from parent after rename

### LINKED ISSUE #276: WinFile screwing up NTFS permissions on Copy
I was RDP into my PC and copy & paste a .dll file across to work with my app. Instead of pasting using the File Explorer as I typically would, I was feeling Nostalgic and used the Paste function in WinFile.

After a bit of troubleshooting why my app wasn't working, I found that there were "Access Denied" errors on that file when using procmon to view it.

This was very strange, especially considering that the app was running as SYSTEM.

I figured out, there is something in the way that WinFile pasted the file which made it give the pasted file explicit permissions which weren't quite right as opposed to just allowing inheriting.

`C:\Test>icacls *.dll
test_paste_using_fileexplorer.dll NT AUTHORITY\SYSTEM:(F)
        BUILTIN\Administrators:(F)
        WIN-JSI09NJG0AO\Administrator:(F)

test_paste_using_winfile.dll NT AUTHORITY\SYSTEM:(I)(F)
         BUILTIN\Administrators:(I)(F)
         BUILTIN\Users:(I)(RX)`

malxau:
NTFS stores ACLs per file.  When a file is moved across directories, ACLs are not changed by the file system.  Providing inherited ACL semantics works in explorer by enabling inheritance on the target after the move, causing its ACL to be updated as a second step.  This change implements that logic.

This is to fix #276 .

### DIFF
--- a/src/wfcopy.c
+++ b/src/wfcopy.c
@@ -3348,10 +3348,28 @@ FileMove(LPTSTR pFrom, LPTSTR pTo, PBOOL pbErrorOnDest, BOOL bSilent)
    *pbErrorOnDest = FALSE;
 
 TryAgain:
-    if (MoveFile((LPTSTR)pFrom, (LPTSTR)pTo))
+    if (MoveFile((LPTSTR)pFrom, (LPTSTR)pTo)) {
+        ACL EmptyAcl;
         result = 0;
-    else
+
+        //
+        //  Enable ACL inheritance, updating the ACL on the file from its new
+        //  parent.  Note this requires Windows 2000 and above.  There's not
+        //  much we can do if it fails, since the rename already occurred.
+        //
+
+        InitializeAcl(&EmptyAcl, sizeof(EmptyAcl), ACL_REVISION);
+        SetNamedSecurityInfo((LPTSTR)pTo,
+                             SE_FILE_OBJECT,
+                             DACL_SECURITY_INFORMATION | UNPROTECTED_DACL_SECURITY_INFORMATION,
+                             NULL,
+                             NULL,
+                             &EmptyAcl,
+                             NULL);
+
+    } else {
         result = GetLastError();
+    }
 
     // try to create the destination if it is not there
 
--- a/src/winfile.h
+++ b/src/winfile.h
@@ -27,6 +27,7 @@
 #include "fmifs.h"
 #include <shellapi.h>
 #include <shlwapi.h>
+#include <aclapi.h>
 #include <strsafe.h>
 #include "suggest.h"
 #include "numfmt.h"


craigwi:
Is this what Explorer does?

malxau:
Yes-ish...do you disagree with my assertion that this is what Explorer does?

For Explorer, see `CFSTransfer::_ResetInheritedACL`.  I see three differences:

1. Explorer queries the ACL on both objects and checks for equality before modifying anything.
2. Explorer inspects the ACL to see if it has inheritance explicitly disabled, and if so, does not reset anything.
3. Explorer uses `TreeSetNamedSecurityInfo` to perform the reset as opposed to `SetNamedSecurityInfo` here.

My 2c would be:

1. It's not clear to me why this compare occurs and why it's an improvement, since this means two queries instead of one modify.  It could be argued that the modify needs to implicitly query the parent, but this seems like a questionable optimization.  Note though that this code runs when renaming a file in a single directory, where there is no reason to change the ACL ever.  I haven't debugged through Explorer to verify that it handles this.
2. This one makes sense to me, I can add it if you think it's beneficial.  That goes partway to addressing the above though, because it implies one of the queries is unavoidable.
3. As far as I can tell from the debugger, `SetNamedSecurityInfo` is still recursive.  This matters because renaming one directory might want to update many children.  `TreeSetNamedSecurityInfo` appears to be a newer API that exists to help Explorer: it does things like supply a progress callback.  The real question here is whether we want to try to display progress while a potentially recursive ACL change occurs.

craigwi:
Thanks @malxau.  I have no reason to contradict your comments and findings.  I do think we should work more to match Explorer if we are going to change anything.

malxau:
In concrete terms then, what change are you requesting?

As I see it, there's two alternatives: an approach that provides progress means this PR gets abandoned and a new approach is needed; changing how to handle the immediate object by querying its current state can be done incrementally here.  Progress is obviously a bigger thing that needs to tie into the message pump, possibly by using a different thread for the ACL change and `PostMessage`-ing progress back to the UI thread.  In general I'm not sure how to think about making larger changes, particularly after making one and it not being merged; I don't want to start down the path of making something large, which becomes difficult to review, and languishes.  If the goal is to do the big thing, that's fine, but it will be big.

craigwi:
My primary concern is that the PR as proposed changes the behavior of Winfile in a significant way, resets ACL on simple rename, and is not the same as Explorer.  I am not suggesting a larger investment in progressive display of any changes, just to clearly understand the result created by Explorer and target that.

So I think the next step, if we proceed here, is to finish the analysis of what Explorer does.

PR #290, opened 2022-01-28
TypeAheadString is buggy in rare situations

### LINKED ISSUE #286: TypeAhead is buggy in rare situations

schinagl:
This solves #286
You can try this out by executing this little .bat file, which creates a bunch of almost empty files
GenerateFiles.zip

schin.jpg
sec1.bla
sec2.jpg
sh.jpg
sim.txt
ssh.ext

If you start typing in e.g. the right dir pane with 's' 's' 'h' the selection lands on sh.jpg instead of ssh.ext
This means if the first two characters are the same, the second is swallowed, and it matches sh.jpg.
This is a bit od, because if you rely on typing you end up on the wrong filename.

This PR fixes #286 and gives Winfile.exe the exact same behavior as Explorer.exe has.

### DIFF
--- a/src/wfutil.c
+++ b/src/wfutil.c
@@ -1569,32 +1569,56 @@ IsBucketFile(LPTSTR lpszPath, PPDOCBUCKET ppBucket)
 // ch == '\0' resets the tick and buffer
 BOOL TypeAheadString(WCHAR ch, LPWSTR szT)
 {
-	static DWORD tick64 = 0;
-	static WCHAR rgchTA[MAXPATHLEN] = { '\0' };
-	DWORD tickT;
-	size_t ich;
+   static DWORD tick64 = 0;
+   static WCHAR rgchTA[MAXPATHLEN + 1] = { '\0' };
+   DWORD tickT;
+   size_t ich;
+   BOOL rVal = FALSE;
 
-	if (ch == '\0')
-	{
-		tick64 = 0;
-		rgchTA[0] = '\0';
-		return FALSE;
-	}
+   if (ch == '\0')
+   {
+      tick64 = 0;
+      rgchTA[0] = '\0';
+      return FALSE;
+   }
 
-	tickT = GetTickCount();
-	ch = (WCHAR)CharUpper((LPWSTR)ch);
-	ich = wcslen(rgchTA);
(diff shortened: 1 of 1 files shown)


craigwi:
When I wrote the original feature for this, the intention was that typing the same character again and again would move to the next item starting with that letter. I like that behavior and your change removes that.

That is, to handle both the current behavior and solve the problem you have, I think you will need to pass in more context or return enough info for the caller to do the matching in a more sophisticated way.

schinagl:
Give it a try, Compile it and run it. 
My change does not remove the behavior, that typing the same letter moves on to the item with that letter. This still works.

Everything works as before except, that the searching is slightly improved: The second character does not get swallowed if they are the same during search

Anyhow you have to compile it and try it out.

craigwi:
Ah, I see.  You are right, the behavior is good.  I suggest a simpler implementation such as:

	tickT = GetTickCount();
	ch = (WCHAR)CharUpper((LPWSTR)ch);
	ich = wcslen(rgchTA);

	// if out of space or more than .5s since last char, start over
	if (tickT - tick64 > 500 || ich > MAXPATHLEN - 2)
		ich = 0;

	rgchTA[ich] = ch;
	rgchTA[ich+1] = '\0';

	tick64 = tickT;

    if (rgchTA[0] == ch) {
        // If one pressed the same character as the first anyhow jump ahead by one
        szT[0] = ch;
        szT[1] = '\0';

        return FALSE;
    }

	lstrcpy(szT, rgchTA);

	return ich != 0;

schinagl:
Perfect! Will take the above one. Thx :-)

PR #291, opened 2022-01-28
ALT + ENTER does not show the properties dialog for root directories

schinagl:
ALT + ENTER does not show the properties dialog for root directories
But one can achieve this anyhow via right mouse on the root directory in the tree control by selecting 'Properties' from the context menu.

This change enables ALT + Enter on the root drive in the tree control aka left pane.

Maybe there is a good reason for the existing behavior, but I don't see the point. Maybe it did not work in older Windows versions, but from my naive point of view everything works nicely.

@craigwi: What is your opinion?

### DIFF
--- a/src/winfile.c
+++ b/src/winfile.c
@@ -1182,9 +1182,8 @@ DefHook:
 // Synopsis: Check if we enable the Properties... menu item
 //           Disable if:
 //
-//           1. The rootdir is selected
-//           2. _ONLY_ the .. dir is sel
-//           3. Nothing is selected in the window with focus
+//           1. _ONLY_ the .. dir is sel
+//           2. Nothing is selected in the window with focus
 //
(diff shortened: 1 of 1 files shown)


craigwi:
The "existing" behavior is in fact the original behavior.  There must have been some limits long ago which are not true now.  I am fine with this change.

malxau:
In the original code, the properties dialog was a hand-rolled dialog displaying time and file attributes.  In FAT this doesn't make sense, because the root directory is a list of files, where each file has those things but the root itself is not described.  It's still interesting to see this behavior though, because NTFS never behaved this way - each file, including the root directory, has an MFT entry describing this information.  So it would always have been valid to display properties for the root directory on NTFS, and it makes sense to do something like change the compressed attribute on the root.

malxau (approved):

PR #294, opened 2022-01-30
Allow right arrow expand without proactively recursively enumerating

### LINKED ISSUE #273: Slowness/Problems with Network Drives
When accessing a network drive on a remote server (i.e. server at a distant site or across internet/VPN), I am seeing slowness drilling down into subdirectories.  It seems that winfile is parsing the entire directory tree when you try to open a subdirectory.  Winfile basically freezes, with the exception of 'Directories Read' in the lower left hand corner, which slowly increments.  When it finishes counting, it un-freezes and works properly.  I've attached a PDF with details.  The original winfile does NOT duplicate this behavior (I tried it via the wfrevive hack).  I've also had trouble with winfile 'hanging' upon opening if you had a window open with a network drive that no longer exists.  You either have to re-map the drive or map another share to the same drive letter in order to get winfile to open.
[PrintSmart Scan.pdf](https://github.com/microsoft/winfile/files/6748980/PrintSmart.Scan.pdf)

malxau:
This is to fix #273 .  Based on the description that this occurs when "Directories Read" is displayed in the status bar and is gradually incrementing, this delay seems highly likely to be caused by `ScanDirLevel`.  This change removes that call unless VIEW_PLUSES is set, and updates the right arrow code to not depend on it.

When looking at the right arrow code, compare to how double click works.  This is handled in `OpenOrEditSelection`, which sends `TC_TOGGLELEVEL` back to `TreeControlWndProc`, which checks the current state, and if it's not expanded, sends `TC_EXPANDLEVEL` to itself.  The code in `TC_EXPANDLEVEL` calls `ExpandLevel` unconditionally whether there is anything known to be under the level or not.  So I believe/assert this must be safe/correct for right arrows too.

Regardless of this change, right arrow into a tree with contents will trigger an enumerate; the previous logic avoids this enumerate if the tree is known to contain nothing, but that means it enumerates the subtree upfront instead.  So I'd also assert that this change means fewer enumerates in total, but it can change when they occur.

### DIFF
--- a/src/treectl.c
+++ b/src/treectl.c
@@ -786,7 +786,7 @@ ReadDirLevel(
                  bResult = FALSE;
                  goto DONE;
               }
-          } else /* if (dwView & VIEW_PLUSES)  ALWAYS DO THIS for arrow-driven expand/collapse */ {
+          } else if (dwView & VIEW_PLUSES) {
              ScanDirLevel(pNode, szPath, dwAttribs & ATTR_HS);
           }
       }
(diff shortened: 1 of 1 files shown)


schinagl:
I also wondered about this behavior and just tried your fix on our 75TB network drive: The 'Directories Read' is gone, and explorer is immediately responsive.
So you have my  :+1:

Issue #270, opened 2021-06-03
Copy/Edit Address


rothgecw:
This isn't so much an issue as an idea for consideration. I was looking for a good way to copy the path at the current directory, but the best I could come up with was to go to properties and copy from there. My simplest suggestion would be to offer an option to copy current path to clipboard. A more complex idea would be to offer an address bar at the top of the main/child window that can be copied or edited (effectively using the Goto Directory functionality).

Thanks so much for bringing Winfile back to life!

SirToffski:
I've found a quick workaround: using <kbd>CTRL</kbd>+<kbd>F</kbd> and copying the path from there.

![image](https://user-images.githubusercontent.com/47118864/145810135-32e04325-9074-4e18-bfec-c5ca3528d86c.png)

schinagl:
Or: Comes already from my spine
ALT + F R 
CTRL + C

SirToffski:
> Or: Comes already from my spine ALT + F R CTRL + C

@schinagl  Neat! Though only a file name is copied with your method, not the path ;)

<kbd>ALT</kbd>+<kbd>F</kbd> <kbd>R</kbd> : file name is copied.

![image](https://user-images.githubusercontent.com/47118864/151176366-c48c8054-0997-4ed1-860e-7ea417e0a9a5.png)

<kbd>CTRL</kbd>+<kbd>F</kbd> : no file name, but the directory the file resides in

![image](https://user-images.githubusercontent.com/47118864/151176624-b0bb9336-1bc7-4089-b66c-069d9be9ab16.png)

Whilst both are useful, @rothgecw was specifically wondering about the path:

> I was looking for a good way to copy the path at the current directory

rothgecw:
@schinagl, @SirToffski, thank you both! Either way seems equally viable, only that the `ALT`+`F` `R` requires the file's parent directory to be selected. I really appreciate both options. Cheers!

schinagl:
I was inprecise... ( Since I really work with Winfile all day, the hotkeys are hardcoded to my brain already :-)

1) You want the full path and you are in the right pane:
ALT + F R
CTRL + C
CTRL + F
CTRL + V
Cursor up
CTRL + C

Full path and filename in copy buffer

This sounds long, but is typed in half a second if you did it a few times

2) You are in the left pane
ALT + F R
CTRL + C

You get the path in the copy buffer

Most times I use 1)

schinagl:
Was this issue solved with #295?
Can we close it?

[...]

PR #295, opened 2022-01-31
When copying, include text to path to enable pasting as strings

malxau:
This is to improve on the situation in #270 .

Since the Windows clipboard is multi-format, copy objects in forms to facilitate pasting (as files) as well as pasting (as text.)

Note this changes how `CheckMultiple` is used.  The existing code would only copy a single _file_ .  This change allows a single _file or directory_ .  I think this makes sense in general and verified that pasting a directory will copy as expected.

### DIFF
--- a/src/wfcomman.c
+++ b/src/wfcomman.c
@@ -1230,29 +1230,41 @@ AppCommandProc(register DWORD id)
    case IDM_CUTTOCLIPBOARD:
 	  {
          LPTSTR  pszFiles;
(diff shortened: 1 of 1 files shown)


schinagl:
Neat :-) Tried this too, works fine in the left and right pane.
Do you think you could also squeeze a line into the Readme.md? Otherwise an nice feature is unnoticed by the average user, who simply downloads things

PR #297, opened 2022-02-01
When junctions are hidden, also hide from left pane

### LINKED ISSUE #268: [Symbolic Links] Please add an option to hide Symbolic Links
Would it be possible in a future update to add a setting into the "By File Type" dialog to hide Symbolic Links please?

I want to see _Hidden_ and _System_ files, but when that setting is turned-on, _Symbolic Links_ also get displayed, which I'd rather not see as accidentally doing a delete or move on them can be destructive.  For example, if a user were to browse into C:/Users/_username_/Documents, they will see "My Music", "My Pictures" and "My Videos".  If they delete or move any of these "folders", the action will occur on the target -- so files could accidentally be destroyed!

Some examples of _Symbolic Link_ "folders" which appear currently:
Winfile_JunctionPointsShown

malxau:
This is to fix #268 .  Although it's not what the title says, looking closer at the screenshots, the concern is around seeing junctions in the left pane (which the user does not have access to), even though the setting for "Show Junction Points" is disabled.

This change leverages `dwReserved0` to identify the reparse tag, which is [now documented behavior](https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-win32_find_dataw) but has existed since reparse points were added in Windows 2000.  I followed the pattern of having the caller filter junctions, which is not optimal - with `dwReserved0` it should be possible to use `dwAttrFilter` for Junctions too, but that requires carefully scrubbing the code to ensure it's propagated.  Most of this change is about propagating it in the left hand pane enumerate.

Also, I'm not stepping back from [my comment](https://github.com/microsoft/winfile/issues/268#issuecomment-1021917860) that the "real" fix is to not have a race condition enumerating the left pane.  That change seems large and complex enough that it's likely to introduce bugs and unlikely to be accepted.  But the advantage of doing it is to eliminate "network hangs" by doing file system operations on the UI thread, which the current code can't achieve.

### DIFF
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -18,7 +18,6 @@
 
 BOOL IsFATName(LPTSTR pName);
 
-
 /* WFFindFirst -
  *
  * returns:
@@ -68,6 +67,14 @@ WFFindFirst(
 	   ATTR_TEMPORARY | ATTR_COMPRESSED | ATTR_ENCRYPTED | ATTR_NOT_INDEXED;
 
    lpFind->fd.dwFileAttributes &= ATTR_USED;
+
+   if (lpFind->fd.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) {
+       if (lpFind->fd.dwReserved0 == IO_REPARSE_TAG_MOUNT_POINT) {
+           lpFind->fd.dwFileAttributes |= ATTR_JUNCTION;
+       } else if (lpFind->fd.dwReserved0 == IO_REPARSE_TAG_SYMLINK) {
+           lpFind->fd.dwFileAttributes |= ATTR_SYMBOLIC;
+       }
+   }
 
    Wow64RevertWow64FsRedirection(oldValue);
 
@@ -139,7 +146,16 @@ WFFindNext(LPLFNDTA lpFind)
          lstrcpy(lpFind->fd.cFileName, lpFind->fd.cAlternateFileName);
       }
 
+      if (lpFind->fd.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) {
+          if (lpFind->fd.dwReserved0 == IO_REPARSE_TAG_MOUNT_POINT) {
+              lpFind->fd.dwFileAttributes |= ATTR_JUNCTION;
+          } else if (lpFind->fd.dwReserved0 == IO_REPARSE_TAG_SYMLINK) {
+              lpFind->fd.dwFileAttributes |= ATTR_SYMBOLIC;
+          }
+      }
+
 	  Wow64RevertWow64FsRedirection(oldValue);
+
       lpFind->err = 0;
       return TRUE;
    }
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -68,14 +68,6 @@ WFFindFirst(
 
    lpFind->fd.dwFileAttributes &= ATTR_USED;
 
-   if (lpFind->fd.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) {
-       if (lpFind->fd.dwReserved0 == IO_REPARSE_TAG_MOUNT_POINT) {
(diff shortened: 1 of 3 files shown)


src/lfn.c
```diff
@@ -18,6 +18,14 @@
 
 BOOL IsFATName(LPTSTR pName);
 
+#ifndef IO_REPARSE_TAG_MOUNT_POINT
```
craigwi, line 21: Why are these defined here?  The values are in elsewhere (e.g., wfdirrd.c) and not defined there.

malxau: No good reason.  I pushed a different commit to remove them.

(The real answer is that my tree is targeting 2000, and symlink in particular was added in Vista, so setting WINVER to anything less than that causes these to not be defined.)

src/lfn.c
```diff
@@ -69,6 +68,14 @@ WFFindFirst(
...
    lpFind->fd.dwFileAttributes &= ATTR_USED;
 
+   if (lpFind->fd.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) {
+       if (lpFind->fd.dwReserved0 == IO_REPARSE_TAG_MOUNT_POINT) {
+           lpFind->fd.dwFileAttributes |= ATTR_JUNCTION;
+       } else if (lpFind->fd.dwReserved0 == IO_REPARSE_TAG_SYMLINK) {
+           lpFind->fd.dwFileAttributes |= ATTR_SYMBOLIC;
+       }
+   }
+
```
craigwi, line 78: This code is a much better way to do what DecodeReparsePoint does.  We should be able to remove one or both calls to DecodeReparsePoint.

src/lfn.c
```diff
@@ -85,8 +92,10 @@ WFFindFirst(
    lpFind->nSpaceLeft = MAXPATHLEN-nLen-1;
 
    if (lpFind->hFindFile != INVALID_HANDLE_VALUE) {
+      DWORD compareAttributes;
+      compareAttributes = lpFind->fd.dwFileAttributes & ATTR_USED;
       lpFind->dwAttrFilter = dwAttrFilter;
```
craigwi, line 97: This is not necessary as lpFind->fd.dwFileAttributes was already combined with ATTR_USED.

src/treectl.c
```diff
@@ -836,7 +862,22 @@ ReadDirLevel(
...
+                  if (lfndta.fd.dwFileAttributes & ATTR_JUNCTION &&
+                      (!(dwAttribs & ATTR_JUNCTION)))
+                  {
+
+                      continue;
+                  }
+              }
+              break;
+          }
       }
```
craigwi, line 881: This block and the one above at line 676 are doing essentially the same thing.  Can we create a method to "SkipJunctionPoints" or something like that?

src/lfn.c
malxau: lpFind->fd.dwFileAttributes are combined with ATTR_USED, then ATTR_JUNCTION/ATTR_SYMBOLIC are added (which are not part of ATTR_USED), then this compare occurs.  The intention was not to compare against ATTR_JUNCTION/ATTR_SYMBOLIC.  Another approach would be to perform this compare, and if the entry is filtered call FindNext (which adds them); if the entry is not filtered then add them here.

src/lfn.c
schinagl: You might want to omit evaluating attributes returned from DecodeReparsePoint(), but you can't remove the call to it, because it delivers szLinkDest, which is needed.

malxau: I changed the one caller that can be changed.

src/lfn.c
malxau: I changed this to the alternate approach which avoids the back-and-forth conversion, effectively setting ATTR_JUNCTION/ATTR_SYMBOLIC after the decision to return the entry has already been made.

src/treectl.c
malxau: Done.

PR #298, opened 2022-02-01
Fix German 'Strg' vs 'Umschalt' mismatch

schinagl:
Add CTRL+N for Connect NetWork Drive
Add CTRL+D for Disconnect NetWork Drive

Add CTRL + B/K/P to File menu since the hot keys existed for a long time (+1 squashed commits)

Dis/Connecting lots of drives during my daily workflow, so I thought this might be useful for other keyboard driven users.
Renamed 'Umschalt' in German .rc to 'Strg'. This was wrong, because 'Umschalt' == Shift. ( I am a german native speaker)
Checked the languages, and they all except German use Ctrl. German uses Strg. Took care of it.


malxau (approved):
I'm now wondering why Ctrl+Shift+N was used to create a directory when Ctrl+N was unused.  Sigh.

craigwi:
Well, in the old days, Ctrl+N was switch to driven N:.

craigwi:
Before moving on with this change, I suggest we map out the potential uses (beyond current ones) of the control+key combinations.  For example, I have a private build which uses ctrl+T to be go to %TEMP%.  I was thinking of using ctrl+D to be goto downloads directory.

schinagl:
I am also fine with CTRL+I for 'DIsconnect Network Drive'

CTRL + T stays for %TEMP%
CTRL + D stays for Download

Its is a different issue, but there are many variables which are useful like %LOCALAPPDATA% %APPDATA% %TEMP% ...
How to handle them?

What about adding 'virtual drives' in 'Select Drive'
grafik
At least one could use ALT + D S <coursor down>*

Anyhow should I change to CTRL + I for this PR?

craigwi:
Thinking about this some more: connect network drive already has a shortcut of sorts: Alt+D,N.  So this would make Ctrl+N == Alt+D,N.   Seems unnecessary or am I missing something.

schinagl:
Following your thoughts would mean we do not need hotkeys at all, because all command can be reached via ALT+<letter for menu> + <key for underlined letter in menu entry>

Anyway. I have reverted the hotkeys for connect and disconnect.
At least lets fix the german 'Strg' vs 'Umschalt' mismatch and document the already existing hotkeys for opening the various shells in the menu.

PR #301, opened 2022-02-03
Display child windows with sunken borders for purely cosmetic reasons

### LINKED ISSUE #242: Add ClientEdge to the MDI childs for better looks
Here one MDI child has ClientEdge, and the other does not

Clipboard_20191107

malxau:
This is to fix #242 .  Note this is cosmetic and subjective, so it's completely fair if people think this not a change we want to make, although if there's a concern about the visual change, I think we should close #242 to clearly indicate this is not something to pursue.  Note the visual change depends a little on the Windows version; I agree with the screenshot in 242 that this works better with the "classic" UI, although it's less clear on a more modern UI that it's visually superior.

In terms of implementation, note that the code has a frame window; within that there's a child which is the MDIClient; and within that are individual windows.  Those windows are created via `WM_MDICREATE` which doesn't specify extended styles.  So this change works by setting extended styles during create, then forcing a recalculation of the client area (since the nonclient area adds an extra pixel by setting this flag.)

### DIFF
--- a/src/wftree.c
+++ b/src/wftree.c
@@ -535,6 +535,7 @@ TreeWndProc(
 
             INT dxSplit;
             DRIVE drive;
+            DWORD dwNewExStyle;
             WCHAR szPath[2 * MAXFILENAMELEN];
 
             //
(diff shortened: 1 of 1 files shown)


schinagl:
Just a general question on look and feel to @malxau
Why don't the childwindows of winfile have the flat design of the main window?
grafik

Would this be possible?

malxau:
> Why don't the childwindows of winfile have the flat design of the main window?

Does this refer to the 3D effect or the window decorations? If it's the 3D effect, that's this change, and I really don't mind reverting it - the only reason for doing it was to close one outstanding issue either as "we've done it" or "we decided not to do it", but don't leave it outstanding.

If it's the window decoration, then I don't think this is practical.  Because winfile is MDI, the child windows are being rendered by GDI but the main window is being rendered by DWM and they're completely different beasts.  In particular, GDI is stuck in Vista Basic because that's when it was assumed DWM would take over everything.  

There's really two options I see, both sucky:
1. Don't use MDI.  Make all windows free floating (similar to explorer), so DWM renders the frame.
2. Manually render our own frame.  This is still using GDI (it's still a bitmap), but it enables the look and feel to move past 2006.  Since today's DWM isn't heavily translucent, it'd probably be possible to mimic fairly closely.  Ideally though this would want to honor the user's choices for active/inactive, including the title bar color, border color, etc - it's not a small amount of work.

schinagl:
No I am fine with the 3d effect, and the sunken border. No need to revert.

The question was about sharp edges, and DWM rendering the childwindows, as I got from your answer.
Going the manually rendered way is way to much effort.
Thx,  Question answered! :-)

OrthodoxWindows:
> 

I see a third possibility: create a personalized visual style, which draws a non-DWM window that looks like the DWM window (a kind of aero-10-basic window). Moreover, the problem of the "Vista-Basic" window, no longer arises when the visual style of Windows 10 goes to Aero Lite.

I think this third possibility and largely the cleanest solution. It is important not to change the operation of the MDI mode.

Issue #149, opened 2018-04-29
Localization needed for Format Drive changes


matthewjustice:
PR #108 / commit ea0dd95 introduced several UI changes that need localization. I'm filing an issue to track this work. Let me know if we want to use a different approach for requesting localization.

Changes needed:
- In res_zh-CN.rc, res_zh-JP.rc, and res_pl-PL.rc 
These strings were already present, but need to be updated to match the new English text shown below:
`MENUITEM "&Format Drive...", IDM_FORMAT`
`MH_MYITEMS+IDM_FORMAT, "Formats a drive"`
This new string needs translation:
`IDS_FORMATSELECTDLGTITLE "Format Drive"`
- In winfile_zh-CN.dlg, winfile_ja-JP.dlg, and winfile_pl-PL.dlg 
This new dialog was added in English and needs localization:
`FORMATSELECTDLG`
- In wfdlgs3.c
Increase the size of CCH_DLG_TITLE if the localized version of IDS_FORMATSELECTDLGTITLE is larger than 16 characters.

FYI for the various people who have been involved with translation: 
@MouriNaruto @mattn @thecatkitty @craigwi

clzls:
In zh-CN translations there seems to be no difference between "Format Drive" and "Format Disk". They are both simply called as "Format" in Windows 10.
So maybe the old translation of "Format Disk" can be left unchanged.

clzls:
> In res_zh-CN.rc, res_zh-JP.rc, and res_pl-PL.rc

A typo. It should be

> ..., res_ja-JP.rc, ...

matthewjustice:
@clzls - Thanks! I corrected the typo in my original post to avoid confusion. I appreciate you taking a holistic look at the appropriate translations.

clzls:
Task status:
- [x] Chinese
- [x] Japanese
- [ ] Polish

clzls:
@mattn ...And the dialog strings don't show up fully in my high DPI PC(Maybe because of my Chinese Windows?). You may extend the width.
![image](https://user-images.githubusercontent.com/16474386/41029617-aea23bb2-69ae-11e8-905b-663b2da78d9d.png)

mattn:
I'll look into it in later.

PR #302, opened 2022-02-04
Save and restore format dialog title rather than duplicating it

malxau:
This is to a fix things noted in #149 .

In the current code, the format select dialog is shown, which invokes `SHFormatDrive`, and that function can change the title of the parent window, so the code restores the title from `IDS_FORMATSELECTDLGTITLE`.  In other words, the dialog's title is recorded twice, once in the dialog resource (used when it is displayed) and once in `IDS_FORMATSELECTDLGTITLE` which is used on the dialog if `SHFormatDrive` fails to complete successfully.

Unsurprisingly, these two strings became out-of-sync as a result of this duplication, despite being intended to be the same string.

This PR does three things:

- Save and restore the title across the call to `SHFormatDrive`, which eliminates the need for `IDS_FORMATSELECTDLGTITLE` so titles are stored once.
- Use a memory allocation rather than `CCH_DLG_TITLE` so it will function regardless of the length of the dialog title.  #149 comments how `CCH_DLG_TITLE` should be reviewed to verify it's large enough for all languages, and it wasn't large enough for de-DE, so on return from `SHFormatDrive` the dialog title was being truncated.
- Use the ja-JP version of `IDS_FORMATSELECTDLGTITLE` as the dialog's title, which had not been localized.  For other languages it seemed preferable to keep the dialog title since that's the obvious one that everybody sees so is more likely to be correct; for ja-JP it clearly wasn't correct.  I'd still appreciate verification from @mattn that this string is correct though.


craigwims:
@malxau-msft, if you can fix the conflict, one of us can merge it.

src/wfdlgs3.c
```diff
@@ -788,6 +788,18 @@ FormatSelectDlgProc(register HWND hDlg, UINT wMsg, WPARAM wParam, LPARAM lParam)
                 // that parent, even if it is hidden.
                 ShowWindow(hDlg, SW_HIDE);
 
+                pszDlgTitle = NULL;
+                dwCchTitleLength = GetWindowTextLength(hDlg);
+                if (dwCchTitleLength > 0)
+                {
+                    dwCchTitleLength++;
+                    pszDlgTitle = malloc(ByteCountOf(dwCchTitleLength));
+                    if (pszDlgTitle != NULL)
```
craigwims, line 797: sorry; just noticed.  The WinFile standard for memory allocation is LocalAlloc().

schinagl:
@malxau: It seems we are just a LocalAlloc() away from Merging....

schinagl:
Even this one could go on ...

PR #306, opened 2022-02-19
Left Pane implodes on dead Reparse Points

schinagl:
### Problem
The left pane implodes to the root drive if one clicks on a dead Reparse Point

### How to test
Download ln.exe from https://schinagl.priv.at/nt/ln/ln64.zip and place it in the same directory as the below .bat file
Start [TestDeadReparseImplode.bat](https://github.com/microsoft/winfile/files/8102923/TestDeadReparseImplode.zip) from an administrative command prompt (or grant yourself the right to create symbolic links) in your e.g. temp directory.

It builds a structure like
```
└───dest
    └───sl
        ├───iLib_dead
        └───iLib_symlink

```
Navigate to x:\dest\sl and double click on iLib_symlink. The left pane does not implode to the root as the original version does.

### Comments to the code
There is one little change in CreateDTABlockWorker(), which deals with the failing WFFindFirst and checks if the item was a dead Reparse Point

### DIFF
--- a/src/wfdirrd.c
+++ b/src/wfdirrd.c
@@ -748,10 +748,7 @@ InvalidDirectory:
             if (hwndTree=HasTreeWindow(hwnd)) {
 
                // Check if it is a Reparse Point
-               lpTemp[0] = '\0';
-               DWORD attr = GetFileAttributes(szPath);
-               lpTemp[0] = CHAR_BACKSLASH;
-               if (attr & ATTR_REPARSE_POINT) {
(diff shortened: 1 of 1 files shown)


malxau (approved):

malxau:
I tried this out and it works for me as described.  I think it's better to use FILE_ATTRIBUTE_REPARSE_POINT here, because if an object is inaccessible it doesn't really matter whether it's a link or not - collapsing back to the root won't help.

PR #308, opened 2022-03-04
Developer mode for symbolic links

schinagl:
### Problem
Winfile fails on copying Symbolic Links in developer mode

### How to test
Start [TestCopyReparsePoint.bat](https://github.com/microsoft/winfile/files/8430378/TestCopyReparsePoint.zip)  from an administrative command prompt in your e.g. temp directory. Make sure you are in developer mode and do not have SE_CREATE_SYMBOLICLINK assigned to the active user.

It builds a structure like
```
├───1
└───sl
    ├───iLib
    │   ├───inc
    │   └───src
    ├───iLib_symlink
    │   ├───inc
    │   └───src
    └───zzz
```

Navigate to x:\sl in Winfile and copy it e.g. via Drag and CTRL to x:\1. You will receive no errors opposite to the original version, which fails on symbolic link creation of x:\sl\iLib_symlink. 

### Comments on the code
WFCopy() and MKDir() check the return code of its respective operation and if it was ERROR_PRIVILEGE_NOT_HELD, it creates the symbolic link itself.
This has the drawback, that alternative data streams and EA records are not copied.
* For EA records I would definitley put this under known limitation, because they are legacy from OS/2
* Alternative streams on symbolic links is deficit, yes.

ACLs are not copied, but this also applies for the CreateDirectoryEx()/CopyFile() case, so ACLs are fine, I think.
Junctions work anyhow fine.

Making this whole EA record and alternative streams copy work is quite some work, I already did it for LinkShellExentsion, so I could contribute the coding, but it is a few hundred lines of code. This then also handles weird cases like sparse alternative streams, EA records, compressed and encrypted files, but I guess we are overshooting here.

I much more hope, that a future version of CreateDirectoryEx() and CopyFileEx() in e.g. Win11 support developer mode, so that this problem is then solved.

In general this implements the simplest of all Reparse Point copy strategies, which I called the 'Splice' method. See also the [explanation of Splice](https://schinagl.priv.at/nt/hardlinkshellext/hardlinkshellext.html#splice) in the LinkShellExtension Docu. 

Winfile does the 'Splice' pattern also for inner and outer reparse points. So this will only work for absolute symbolic links. Relative symbolic links will be copied but point to 'something'. This can not be solved with this PR for Winfile, because it needs more sophisticated symlink resolving.

Maybe in a later PR we could adress relative symbolic links.

In the course of action DecodeReparsepoint() was adapted a little, because the second argument was never used, and in my opinion one should pass a path to a reparse point and not path\\\*.* as the first argument

So ... lets see what you guys think.


schinagl:
Maybe my description above was too long or confusing, but basically this is a **bug fix**:
Currently one can not copy a tree with symbolic links without error boxes, when in developer mode.

So maybe @malxau you can give it a look/try? It would bring us forward... :-)

craigwims:
@schinagl, I must be dense.  What do you mean by "developer mode"?

schinagl:
### Problem
The privilege to create Symbolic Links (SeCreateSymbolicLink) is normally only held by administrators.
But 'normal' users also want to create symbolic links, and do not want to enter the admin password each time via the UAC dialog.

### So there are two ways out of this
a) Enable 'developer mode': https://www.howtogeek.com/292914/what-is-developer-mode-in-windows-10/
b) Grant the privilege to create symbolic links in the policy editor to a specific user: https://schinagl.priv.at/nt/hardlinkshellext/hardlinkshellext.html#changesymboliclinkprivilege

With a) or b) enabled a normal user can create Symbolic Links

### What is the point of this fix
1. We use CopyFileEx instead of CopyFile, and ask it to copy Symbolic links if possible. ( .... COPY_FILE_COPY_SYMLINK) ( Files are copied normally anyhow ... ) [CopyFileEx instead of CopyFile, because in the good case (== SeCreateSymbolicLink held) anyhow it copies the symlink and we are done]
2. In case of failure CopyFileEx will return with ERROR_PRIVILEGE_NOT_HELD, which means we are not allowed to create symbolic links. ( See above for the ways out )
3. Then we give it another try and try to create the Symbolic Link set ourselves with SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE flag set to CreateSymbolicLink().
4. Assuming the user had 'developer mode' on, we are now able to create a symbolic link and saved the day:-)

Basically we do similar things for CreaDirectoryEx() and also here provide a fallback to create symbolic links if 'developer mode' is on

craigwims:
thanks!  Which version of Windows introduced CopyFileEx?

schinagl:
Windows XP/Windows Server 2003

src/wfdirrd.c
```diff
@@ -1209,11 +1193,20 @@ DWORD DecodeReparsePoint(LPCWSTR szMyFile, LPCWSTR szChild, LPWSTR szDest, DWORD
...
+			// Handle ?\ prefix
 			if (szT[0] == '?' && szT[1] == '\\')
 			{
 				szT += 2;
 				cwcLink -= 2;
 			}
+			else
+				// Handle \??\ prefix
+				if (szT[0] == '\\' && szT[1] == '?')
+				{
```
craigwims, line 1206: Let's check all four characters.

src/wffile.c
```diff
@@ -105,6 +105,12 @@ OpenFileForCompress(
     PHANDLE phFile,
     LPTSTR szFile);
 
+DWORD
+WFCopySymlink(
+   LPTSTR pszFrom,
+   LPTSTR pszTo,
+   DWORD dwFlags,
+   DWORD dwNotification);
```
craigwims, line 114: This should go in winfile.h; the other declarations above this are for functions in this file.

src/lfn.c
```diff
@@ -482,6 +483,28 @@ LFNMergePath(LPTSTR lpMask, LPTSTR lpFile)
    return TRUE;
 }
 
+/* WFCopySymlink
+ *
+ *  Copies symbolic links 
+ */
+DWORD
+WFCopySymlink(LPTSTR pszFrom, LPTSTR pszTo, DWORD dwFlags, DWORD dwNotification)
+{
```
craigwims, line 492: This function only 'copies szFrom to szTo' if pszFrom is a symbolic link.  I think the name should include the conditionality; e.g., WFCopyIfSymlink.

schinagl:
Added Review fiindings

schinagl:
A few good questions, lets start with 
### How do I test the case of WFCopyIfSymlink in lfn.c?

1. Unfortunately my published testcase lacked a symbolic link file. Sorry. I have updated [my testcase](https://github.com/microsoft/winfile/files/8430378/TestCopyReparsePoint.zip) so that it also contains a symbolic link file

2. The only way to test this is to 
- enable developer mode as admin
- create a non admin user
- log on as non-admin
- put a breakpoint to WFCopy:548

and copy the given tree from sl/ to 1/ as described above

Then it looks like this
image

It is important, that your Winfile, does not **hold** SeCreateSymbolicLinkPrivilege. (Use procexp.exe!) You **must not** see it in the list of privileges. If you see it and it is 'Disabled' then you hold the privilege and it can not be reproduced.

### When I refresh by expand/collapse 'sl', the icon is correct.
Can reproduce. Have to check. In my original test this worked, but we changed a few things in #309. Have to see what is the reason.

### If I create a directory '2' next to '1' and ctrl+drag 'iLib_symlink' to '2
Can reproduce. This is odd, Will also investigate. I have a rough idea

In general sorry for the inconvenience. Will do by best to get it fixed

schinagl:
Fixed 'If I create a directory '2' next to '1' and ctrl+drag 'iLib_symlink' to '2'

schinagl:
Fixed 'When I refresh by expand/collapse 'sl', the icon is correct.'

craigwims:
The issues are fixed; thanks!

I do have one concern about a code change; see the new comment.

src/wfcopy.c
```diff
@@ -1699,10 +1699,11 @@ GetNextPair(PCOPYROOT pcr, LPTSTR pFrom,
...
 
-			   //
-               // Directory: operation is recursive.
+			      //
+               // Directory: operation is recursive, but not for junctions and symlinks
                //
-               pcr->fRecurse = TRUE;
+               if (!(pDTA->fd.dwFileAttributes & (ATTR_SYMBOLIC | ATTR_JUNCTION)))
+                  pcr->fRecurse = TRUE;
                pcr->cDepth = 1;
```
craigwims, line 1707: this code changes from always setting fRecurse to TRUE to sometimes setting it to TRUE.  Code just above sets that field to FALSE.  Thus I don't think we can be sure the field is FALSE and thus only have to conditionally set it to true.

So, seems like we should set fRecurse either way.  Make sense?

schinagl: Right. Now it is set correctly regardless of any precondition

PR #310, opened 2022-03-11
Pressing CTRL + Enter runs the assoicated file as Administrator

schinagl:
Pressing CTRL + Enter runs the assoicated file as Administrator


src/res.rc
```diff
@@ -38,6 +38,7 @@ BEGIN
     VK_ESCAPE,  IDM_ESCAPE,     NOINVERT, VIRTKEY
     VK_RETURN,  IDM_OPEN,       NOINVERT, VIRTKEY
     VK_RETURN,  IDM_OPEN,       NOINVERT, VIRTKEY, SHIFT
+    VK_RETURN,  IDM_OPEN,       NOINVERT, VIRTKEY, CONTROL
```
malxau, line 41: Can I ask why we do this? I know this PR is just following a pattern, but it looks like this causes User32 to match the Control+Return combination, then map it to the same ID as other keystrokes, then Winfile peeks at whether the user is currently holding down Control to determine which action to take, rather than having a unique ID that makes this unambiguous.

Winfile is checking the state of Control when processing the message, not when the message is sent, so it becomes possible to do things like Press Enter, Release Enter, Press Control, and have Winfile assume it's Control+Enter.

src/wfcomman.c
```diff
@@ -714,7 +714,7 @@ OpenOrEditSelection(HWND hwndActive, BOOL fEdit)
       }
       else
       {
-          ret = ExecProgram(szPath, szNULL, NULL, (GetKeyState(VK_SHIFT) < 0), FALSE);
+          ret = ExecProgram(szPath, szNULL, NULL, (GetKeyState(VK_SHIFT) < 0), (GetKeyState(VK_CONTROL) < 0));
```
malxau, line 717: I assume you decided there's no point having this check in the notepad/editor/F12 cases above? "Edit file as admin" seems more useful IMHO to "open the file in a minimized editor".

malxau (approved):

schinagl:
If you omit the very line in res.rc then you get no notification with IDM_OPEN, when pressing CTRL + ENTER

If you add an extra ID in res.rc, eg, IDM_ADMIN_OPEN, then you have to catch it in wfcomman.c:948 and pass the 'admin-fact' through OpenOrEditSelection(), and thus change also signatures of OpenOrEditSelection() (which is used from other places).

So long story short: The motivation to my solution was having a minimal invasive change, which follows a common pattern in this project.

src/wfcomman.c
schinagl: I didn't need 'Edit as admin'.
I needed to run bat files as admin, the same way as Explorer does with 'Run As Administrator'.

But you are right we could also do this for fEdit == TRUE, to have it 'symmetrical'

malxau:
Right, on reflection I think the issue is having multiple IDs implies having multiple menu entries.  The current code allows a single menu entry to do different things based on the keyboard modifiers.  I'm still a bit torn about that, because having one menu entry do different things is not discoverable; on the other hand the current menus have lots of options already.

As far as being minimally invasive, I'm hoping once we're more familiar with the code we can make larger changes while still being able to review them and not introduce additional risk.  Maybe that's optimistic...

malxau (approved):

PR #315, opened 2022-04-13
Symlink Icons are not shown after (un)compress

schinagl:
### Problem
In the aftermath of #308 I stumbled about one detail we missed:
After (un)compress on a symbolic link the symlink icon is not shown anymore.

### How to test
Start [TestCopyReparsePoint.bat](https://github.com/microsoft/winfile/files/8430378/TestCopyReparsePoint.zip)  from an administrative command prompt in your e.g. temp directory. 

It builds a structure like
```
├───1
└───sl
    ├───iLib
    │   ├───inc
    │   └───src
    ├───iLib_symlink
    │   ├───inc
    │   └───src
    └───zzz
```

Select sl/iLib_symlink and choose from the menu File/(Un)Compress. The symlink icon of sl/iLib_symlink is shown properly contrary to the original. 

### Comments on the code
There are two locations in the source where the attributes are fetched directly via GetFileAttributes() and thus bypass WFFind{First|Next} extended in #297
1. treectl.c/InsertDirectory()
2. wffile.c/RedrawAllTreeWindows()

The topic 1. was handled in #308 as a byproduct, but 2. is addressed in this PR. 
Anyhow, the fix is simple and introduces a helper function SetNodeAttribs(), which is called from 1) and 2)

### DIFF
--- a/src/treectl.c
+++ b/src/treectl.c
@@ -153,6 +153,36 @@ GetTreePath(PDNODE pNode, register LPTSTR szDest)
 
 
 
+//  SetNodeAttribs
+//                              
+//  Set node attributes for directory/junction/symlink
+//
+VOID
+SetNodeAttribs(PDNODE pNode, LPTSTR szPath)
+{
+   pNode->dwAttribs = GetFileAttributes(szPath);
+   if (INVALID_FILE_ATTRIBUTES == pNode->dwAttribs) {
+      pNode->dwAttribs = 0;
+   }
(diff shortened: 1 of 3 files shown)


craigwims:
Glad you found this.  It is one of the perils of working on an old code base with some redundancy "built" in -- e.g. #define ATTR_DIR            FILE_ATTRIBUTE_DIRECTORY.

PR #316, opened 2022-04-13
Wrong position of selection after copy

schinagl:
### Problem
When copying a tree above a selection the selection is way too below after the copy 
See [this short video](https://user-images.githubusercontent.com/7418603/163217428-fbf7e5cb-6cde-4467-b36a-593def056ce8.mp4)

The selection after copy is on 'IterateLPXDTA' which is wrong. It should be 'sl'. I checked a one year old version of Winfile and it is also there. So no side-effect of recent changes

### How to test
Start [TestCopyReparsePoint.bat](https://github.com/microsoft/winfile/files/8430378/TestCopyReparsePoint.zip)  from an administrative command prompt in your e.g. temp directory.

It builds a structure like
```
├───1
└───sl
    ├───iLib
    │   ├───inc
    │   └───src
    ├───iLib_symlink
    │   ├───inc
    │   └───src
    └───zzz
```

Select sl/iLib_symlink and copy to 1/ and check the location after copy. 
It should [look like this](https://user-images.githubusercontent.com/7418603/163217725-2e975350-00f4-40e3-a3df-c480cf30e614.mp4)

### Comments on the code
It seems that there is code in treeCtl.c:~2470 which is superfluous, because some code else already does the repositioning after copy

### DIFF
--- a/src/treectl.c
+++ b/src/treectl.c
@@ -2452,15 +2452,6 @@ TreeControlWndProc(
             }
          }
 
-         //
-         // if we are inserting before or at the current selection
-         // push the current selection down
-         //
-         nIndex = (INT)SendMessage(hwndLB, LB_GETCURSEL, 0, 0L);
-         if ((INT)LOWORD(dwTemp) <= nIndex) {
-            SendMessage(hwndLB, LB_SETCURSEL, nIndex + 1, 0L);
-         }
-
          break;
 
       case FSC_RMDIR:


craigwims:
These lines were in the original sources I found.  I think we should try to track down the case for which they did their job and not just delete them.

schinagl:
The intention of these lines is obvious, and once they were written they were correct. 

Nowadays it seems that some code else pushes down the selection and thus they are not needed anymore. 

Deleting them does fix the issue. If you have an other explanation I am eager for it.

malxau:
(Sorry for being MIA.)

This seems really odd to me since AFAIK LB_INSERTSTRING moves the selection automatically.  I added some debug prints to verify my sanity, and sure enough, I am sane (at this specific moment.)

Because this behavior is so longstanding, working with this source didn't seem likely to find what happened, so I started looking at the binaries in old releases.  NT 4 has this bug; NT 3.51 does not.  The relevant code snippet was added to winfile way back in 1991 in a commit called "win31 merge".

The next thing I tried was running winfile from NT 3.51 on NT 4.  Since these were inbox, it initially complains and terminates due to an unresolved external, but with some debug magic, I got it to work.  And the bug is present.

Similarly, the bug is also present in Windows Me (the only build from that lineage I have running right now.)  Also, I can take winfile.exe from Windows 3.1 (the 16 bit one) and observe the same bug in NT 4.0.

So it really looks like the behavior of LB_INSERTSTRING changed in 4.0, and it applies the changed behavior even to 3.x and 16 bit processes.

Given that, I'd agree with @schinagl that this code made sense when it was written, but hasn't made sense for 25 years.

craigwims:
Thanks @schinagl and @malxau for the analysis; just the sort of thing we need to be confident about this sort of change.

PR #312, opened 2022-03-15
Creation of Symbolic Links, Hardlinks and Junctions

schinagl:
### General
Creation of Symbolic Links, Hardlink and Junctions via keyboard modifiers

Pressing key combinations during moving items with the mouse enables
CTRL + SHIFT for Symbolic Links on files or directories
CTRL + SHIFT + ALT for Hardlinks on files and Junctions on directories
File_Manager_symlink

### Meta Thoughts
I know of the constraints on bigger PRs... ;-) They might not make it to the maintrunk.
Anway I needed it. And I think I was very careful when doing changes.
So I hope in your patience when reviewing it.

### Comments on the code
All starts in treectl.c and wfdirrd.c, where the drag loop is modified to recognize when different key modifiers are pressed. 

Upon these modifiers fShowSourceBitmaps is set to one of 5 choices
1 == TRUE  == DROP_COPY
0 == FALSE == DROP_MOVE
2 ==       == DROP_LINK
3 ==       == DROP_HARD
4 ==       == DROP_JUNC
where DROP_COPY and DROP_MOVE intentionally match the old values TRUE/FALSE when only copy and move was possible

DMMoveCopyHelper() was modified to translate the above new 'operators' into an extended set of functions
FUNC_LINK       0x0008
FUNC_HARD       0x0009
FUNC_JUNC       0x000A
for WFMoveCopyDriverThread()

WFMoveCopyDriverThread() is modified to process the new commands

There was bit-magic in GetMoveCopyCursor(). Extended this accordingly for DROP_LINK, DROP_HARD, DROP_JUNC

A few messages were added and all are translated into all existing languages properly. (You can use deepl or google translate to verify)

4 more Cursors were added.

Updated the documentation

### About Testing
Most of the new code is aside the existing code. 
E.g. WFMoveCopyDriverThread() contains a few more case statements in the switch statements, which are only used with new functionality, so there should be little impact.
I checked all the 4 places where DMMoveCopyHelper() is used, and it is compatible, because DROP_COPY == TRUE and DROP_MOVE == FALSE.

Tried out with confirmation enabled.

I had Winfile running during the last 2 weeks on my productive system and no issues occured. ( I really use hardlinks during my daily routine work ... )

Tried to stick with the very local coding convention, so the changes are minimal invasive as possible.

PR #318, opened 2022-04-26
Netshare/NetUnShare Button always grayed out

schinagl:
### Problem
Net Share not working. The share icon in the toolbar was always grayed out
netsharegrayedout

### Comments on the code
Winfile used an API to share network resources from ntlanman.dll, which is not available anymore with Windows7/10
Changed it to use ShowShareFolderUI() from Ntshrui.dll. Works. 

Unsharing is the problem. How does one open this dialog
StopShare

Which is invoked through explorer via
GiveAccess

Any ideas? @craigwi , @malxau can you have a look behind the scenes on how explorer does this? Is it in Shell32.dll ? Or where is it?
I was not able to google this at first hand, maybe you are faster on this

### DIFF
--- a/src/wfcomman.c
+++ b/src/wfcomman.c
@@ -1667,9 +1667,12 @@ AppCommandProc(register DWORD id)
 
       if (lpfnShowShareFolderUI) {
          LPTSTR szDir;
-         BOOL bDir = FALSE;
+         HRESULT hr;
+         BOOL bDir;
+            
+         bDir = FALSE;
          szDir = GetSelection(1 | 4 | 16, &bDir);
-         HRESULT hr = ShowShareFolderUI(hwndFrame, szDir);
+         hr = ShowShareFolderUI(hwndFrame, szDir);
          if (hr != S_OK) {
             FormatError(TRUE, szMessage, COUNTOF(szMessage), ERROR_INVALID_SHARENAME);
 
--- a/src/wfcomman.c
+++ b/src/wfcomman.c
@@ -1665,52 +1665,26 @@ AppCommandProc(register DWORD id)
       //
       WAITNET();
 
-      if (lpfnShareStop) {
+      if (lpfnShowShareFolderUI) {
+         LPTSTR szDir;
+         BOOL bDir = FALSE;
+         szDir = GetSelection(1 | 4 | 16, &bDir);
+         HRESULT hr = ShowShareFolderUI(hwndFrame, szDir);
+         if (hr != S_OK) {
+            FormatError(TRUE, szMessage, COUNTOF(szMessage), ERROR_INVALID_SHARENAME);
 
-         if (ret = ShareCreate(hwndFrame)) {
-
-            //
-            // Report error
-            // ret must the error code
-            //
-            goto DealWithNetError;
-
-         } else {
-
-            InvalidateAllNetTypes();
+            LoadString(hAppInstance, IDS_NETERR, szTitle, COUNTOF(szTitle));
+            MessageBox(hwndFrame, szMessage, szTitle, MB_OK | MB_ICONSTOP);
          }
+         InvalidateAllNetTypes();
       }
-
       break;
 
    case IDM_STOPSHARE:
 
       //
-      // Check to see if our delayed loading has finished.
+      // IDM_STOPSHARE not shown anymore, because there is no way to open then 'Stop Share Dialog' with W7/10/11 
       //
-      WAITNET();
-
-      if (lpfnShareStop) {
-
-         if (ret = ShareStop(hwndFrame)) {
-
-            //
-            // Report error
-            // ret must the error code
-            //
-
-DealWithNetError:
-
-            FormatError(TRUE, szMessage, COUNTOF(szMessage), ret);
-
-            LoadString(hAppInstance, IDS_NETERR, szTitle, COUNTOF(szTitle));
-            MessageBox(hwndFrame, szMessage, szTitle, MB_OK|MB_ICONSTOP);
-
-         } else {
-
-            InvalidateAllNetTypes();
-         }
-      }
       break;
 
 #if 0
--- a/src/wfinfo.c
+++ b/src/wfinfo.c
@@ -1736,17 +1736,17 @@ NetLoad(VOID)
 
    if (WNetStat(NS_SHAREDLG)) {
 
-      hNTLanman = LoadLibrary(NTLANMAN_DLL);
+      hNtshrui = LoadLibrary(NTSHRUI_DLL);
 
-      if (hNTLanman) {
+      if (hNtshrui) {
+         lpfnShowShareFolderUI = (PVOID)GetProcAddress(hNtshrui, "ShowShareFolderUI");
+         if (lpfnShowShareFolderUI)
+            PostMessage(hwndToolbar, TB_ENABLEBUTTON, IDM_SHAREAS, TRUE);
+         else {
 
-#define GET_PROC(x) \
-         if (!(lpfn##x = (PVOID) GetProcAddress(hNTLanman, NETWORK_##x))) \
-            goto Fail
-
-         GET_PROC(ShareCreate);
-         GET_PROC(ShareStop);
-#undef GET_PROC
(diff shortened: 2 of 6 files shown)


malxau:
That interface doesn't seem to be documented.  Ntshrui.dll exposes COM interfaces that are only addressable by GUID, and the GUIDs aren't described in SDK headers or documentation.

That said, isn't the higher level issue that winfile is probably not running elevated, and these commands require elevation, so implementing them requires a helper process? Whether the DLL exports are present or not, they can only succeed if the explorer infrastructure exists on the machine?

schinagl:
This is not an elevation issue, because
- Winfile on the _netshare branch already can successfully provide shares via ntshrui.dll
- Explorer can stop sharing via 'Right Mouse/Give Access/Remove access' without elevation. 

Undocumented most likely yes. 

But which coding is behind 'Right Mouse/Give Access/Remove access'
- Is it a COM call to an interface provided by ntshrui?
- Or is it a (COM)call to something else?

I guess it is something else, because the very strings on this dialog can not be found in the resources of ntshrui.dll

But what?

Only the source knows.

schinagl:
Ntshrui.dll is a COM Server, and it supports about 16 interfaces. Thats what you find, when you search the registry.
None of the 16 GUIDs is documented.
Is there a way to ask Microsoft officially if the share stop dialog is one of this 16 interfaces? Or how to use them?
Or? I am stuck here. I need you help

schinagl:
Finally I gave up on the 'Stop Share Dialog', because it is definitley not available anymore for non MS components or outside of MS.
It most likley is called via COM on ntshrui.dll and shows this stacktracke when in use
StopShareStackTrace
Neither ntrshui.dll::CreateWizard nor class CSharePageSharedData is document in any place I could find
So sorry, even if this is a MS component I do not have access to docu on this

If at any time a solution to showing the dialog comes up it can be easily enabled. I have documented all places

For now lets roll with only the 'Share' button and 'Share As...'  menu entry

schinagl:
Since #312 needs more time, maybe we can do a simple one in the meantime?
This one is on simply getting rid of old stuff, and using the proper dialog for sharing.

We know that this branch is dedicated to W7/W8/W10/W11/... so there should be nothing wrong with these changes. 
What do you think?

src/wfcomman.c
```diff
@@ -1665,52 +1665,26 @@ AppCommandProc(register DWORD id)
...
-            // ret must the error code
-            //
-            goto DealWithNetError;
-
-         } else {
+      if (lpfnShowShareFolderUI) {
+         LPTSTR szDir;
+         BOOL bDir = FALSE;
+         szDir = GetSelection(1 | 4 | 16, &bDir);
+         HRESULT hr = ShowShareFolderUI(hwndFrame, szDir);
```
malxau-msft, line 1672: Please avoid C99 variable declarations.

(I know, I'm not trying to comment on the quality of the language construct.  It's just that the newest compiler that targets Windows 2000 doesn't support them, so I end up going through and replacing them all with C89.  This isn't really because Windows 2000 is an interesting target, it's because when somebody points out unexpected behavior I'm going through old releases to try to figure out what changed, when, and why.)

malxau-msft, line 1672: Please avoid C99 variable declarations.

(I know, I'm not trying to comment on the quality of the language construct.  It's just that the newest compiler that targets Windows 2000 doesn't support them, so I end up going through and replacing them all with C89.  This isn't really because Windows 2000 is an interesting target, it's because when somebody points out unexpected behavior I'm going through old releases to try to figure out what changed, when, and why.)

src/winfile.h
```diff
@@ -1037,8 +1037,8 @@ Extern DWORD (CALLBACK *lpfnWNetFormatNetworkNameW)(
                     DWORD    dwFlags,
                     DWORD    dwAveCharPerLine
                     );
-Extern DWORD (CALLBACK *lpfnShareCreate)(HWND);
-Extern DWORD (CALLBACK *lpfnShareStop)(HWND);
+Extern DWORD (CALLBACK *lpfnShowShareFolderUI)(HWND, LPWSTR);
+Extern DWORD (CALLBACK *lpfnShareStop)(LPCWSTR, LPCWSTR, DWORD, HWND);
```
malxau-msft, line 1041: Shouldn't this be removed?

malxau-msft, line 1041: Shouldn't this be removed? Along with NETWORK_ShareStop?

src/wfinfo.c
```diff
@@ -1756,22 +1756,10 @@ NetLoad(VOID)
...
 
-      } else {
-Fail:
-         //
-         // Disable the share buttons/menus
-         // Since WNetStat(NS_SHAREDLG) ret'd true, then we added the
-         // buttons.  Mistake.  Disable them now.
-         //
-         PostMessage(hwndToolbar, TB_ENABLEBUTTON, IDM_SHAREAS, FALSE);
          PostMessage(hwndToolbar, TB_ENABLEBUTTON, IDM_STOPSHARE, FALSE);
```
malxau-msft, line 1759: Shouldn't this be removed? If I'm reading this right, the toolbar will never have an IDM_STOPSHARE item so this will always fail?

malxau-msft, line 1759: Shouldn't this be removed? If I'm reading this right, the toolbar will never have an IDM_STOPSHARE item so this will always fail?

src/wfinfo.c
```diff
@@ -1756,22 +1756,10 @@ NetLoad(VOID)
...
-
-         EnableMenuItem(GetMenu(hwndFrame), IDM_SHAREAS,
-            MF_BYCOMMAND | MF_GRAYED );
-
-         EnableMenuItem(GetMenu(hwndFrame), IDM_STOPSHARE,
-            MF_BYCOMMAND | MF_GRAYED );
+         EnableMenuItem(GetMenu(hwndFrame), IDM_STOPSHARE, MF_BYCOMMAND | MF_GRAYED);
```
malxau-msft, line 1760: As with toolbar, isn't this never inserted anymore so this item is never present?

More generally, is there a reason for not removing IDM_STOPSHARE altogether?

malxau-msft, line 1760: As with toolbar, isn't this never inserted anymore so this item is never present?

More generally, is there a reason for not removing IDM_STOPSHARE altogether?

malxau-msft (changes requested):

src/wfinfo.c
schinagl: I am not sure.
What I got from the code one can have the toolbar configured. This configuration is stored in the registry and is restored upon startup.

So it could happen, that one has configured toolbar with StopShare (even if it is useless), and it would be enabled if we omit this. The struct in tbar.c:55 is only an initial state

This is the same reason why we can not cleanup the bitmaps with toolbar icons.

Anyhow I might be mistaken, and that's why I kept it in. Hmm?

src/wfinfo.c
schinagl: Same answer as on the toolbar icons

src/wfinfo.c
schinagl: Dupe comment

src/wfinfo.c
schinagl: Duplicate comment

src/winfile.h
schinagl: Duplicate comment

src/wfcomman.c
schinagl: Duplicate comment

schinagl:
Updated with code review changes

src/wfinfo.c
malxau-msft: Thanks for the explanation, makes sense.  I'd never looked at the toolbar customize case before.  Looking now though it seems to have some strange behaviors:

- If a toolbar button is already displayed (enabled or disabled), the user can configure its location
- Doing so will cause a disabled button to become enabled
- If the button is not displayed, it is only available to add if it would be enabled (see `HIDEIT`.)

The second point seems like a bug and the third seems very intentional but also strange.  It means that if somebody had customized their toolbar, they'll have stop sharing on it forever - but we can't add it for testing, because it's always disabled and hence can't be added.

PR #319, opened 2022-04-27
Search stops with error when it comes along a broken reparse point.

schinagl:
### Problem
Search stops with error when it comes along an illegal reparse point. (This is not a dangling reparse point, but an illegal!)
An illegal reparse point contains illegal reparse data.

### Comments on the code
Basically it is an extension of the error handling in two places, which deals with error codes that come from illegal reparse data.
I accidentally stumbled across such a situation, when the search for a file stopped with the message 'The data present in the reparse point buffer is invalid.` This is ERROR_INVALID_REPARSE_DATA

Then I started to create illegal reparse points via the debugger with an extra tool, and found ERROR_CANT_ACCESS_FILE to be malicious too. Both are handled now.

### DIFF
--- a/src/wfdirrd.c
+++ b/src/wfdirrd.c
@@ -710,7 +710,9 @@ CreateDTABlockWorker(
       //
       // Try again!  But first, see if the directory was invalid!
       //
-      if (ERROR_PATH_NOT_FOUND == lfndta.err) {
+      if (ERROR_PATH_NOT_FOUND == lfndta.err ||
+         ERROR_INVALID_REPARSE_DATA == lfndta.err ||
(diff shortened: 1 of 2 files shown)


malxau-msft:
You probably saw this; it also means selecting a bad reparse point will succeed with "The specified path is invalid" in the right hand pane.  The old code would display a dialog and refuse to select the folder.  I tried a few other operations and they appear to fail gracefully enough.

schinagl:
I know this, but a dialog is totally unusable during daily work. 

Furthermore this dialog sometimes freezes the whole winfile. Don't ask me why.

So the point was to end up in 'invalid path specified' for all strange situations and have no extra confirmation.

PR #322, opened 2022-05-09
Fix Merge error

schinagl:


malxau-msft (approved):

malxau-msft:
Thanks for promptly checking and following up on this.  I'm reflecting on how we could have prevented it - do you have suggestions?  (There always seems to be a general problem in code reviews where changes are made that aren't as thoroughly tested as the original, so code at the end of a review is worse than at the start...)

schinagl:
It was my bad. I was so sure, everything was merged, and didn't run it again..... and it was late yesterday

The general problem with Winfile is, that it lacks an automated regression test. But this will not be solved here anymore, so we have to be *very* careful.

PR #323, opened 2022-05-12
Wrong selection when typing _

schinagl:
### Problem
The directory pane suffers from some weird behavior of the list box. If you press _ it selects from anchor to caret:

- Run [wrong_selection.zip](https://github.com/microsoft/winfile/files/8682133/wrong_selection.zip). It creates a few almost empty files in a directory.
wrong_selection
- Select the file installservices.exe
- Start typing 'i' 'n' 't' '\_'
You will end up with this
bad_selection

But all you wanted was to type 'int_' have 'int_bla.dll' selected

### Comments on the code
Well this was hard to find out, but it seems the used listbox in Winfile from comctl32.dll has a virtual key behind '\_', which selects all items from anchor to caret.
In wfdir.c:603 @ WM_CHARTOITEM at the return of the 'case', the actual position is returned with respect to the typed character.
If e.g. anchor = 2, and caret = 2, and this case statements returns 5, because the item with '\_' was on the 5th position, then internally the caret is set to 5. This works perfectly as long the typed character is != '\_'.
If it was '\_' the comctl32.dll magic happens, and it selects from anchor (2) to recently returned and set caret (5).
But you don't want this.

To solve this we check for the magic '\_' character and then select it via LB_SETSEL and thus correct anchor and caret before.

It is definitely comctl32.dll which causes this unexpected selection, because I debugged this and it happens within Sel_BlockHilite() inside comctl32.dll. If this is a bug or a feature, I don't know. 

I also checked with a very old version on this branch, and this behavior is in ever since.

### DIFF
--- a/src/wfdir.c
+++ b/src/wfdir.c
@@ -646,8 +646,12 @@ DirWndProc(
          return -2L;
 
       int pos = (i + j) % cItems;
-      if ('_' == LOWORD(wParam))
-         SendMessage(hwndLB, LB_SETSEL, 1, pos);
+
+      // There is a weird behavior in listbox which selects all between anchor an caret
+      // if SHIFT is pressed. Since we return the position here and thus caret will be 
+      // updated, anchor is behind, and pressing shift selects all between anchor and caret
+      // To overcome this we select the current position, and bring anchor and cart in sync.
+      SendMessage(hwndLB, LB_SETSEL, 1, pos);
 
       return pos;
    }
--- a/src/wfdir.c
+++ b/src/wfdir.c
@@ -608,6 +608,7 @@ DirWndProc(
       LPWSTR szItem;
       WCHAR rgchMatch[MAXPATHLEN];
       SIZE_T cchMatch;
+      UINT pos;
 
       if ((ch = LOWORD(wParam)) <= CHAR_SPACE || !GetWindowLongPtr(hwnd, GWL_HDTA))
          return(-1L);
@@ -645,7 +646,7 @@ DirWndProc(
       if (j == cItems)
          return -2L;
 
-      int pos = (i + j) % cItems;
+      pos = (i + j) % cItems;
 
       // There is a weird behavior in listbox which selects all between anchor an caret
       // if SHIFT is pressed. Since we return the position here and thus caret will be 
--- a/src/wfdir.c
+++ b/src/wfdir.c
@@ -645,7 +645,11 @@ DirWndProc(
       if (j == cItems)
          return -2L;
 
-      return((i + j) % cItems);
+      int pos = (i + j) % cItems;
+      if ('_' == LOWORD(wParam))
+         SendMessage(hwndLB, LB_SETSEL, 1, pos);
+
+      return pos;
    }
    case WM_COMPAREITEM:
 


craigwims:
wow.  That is weird.  @malxau-msft, can you confirm and also surface any other hidden behavior?

schinagl:
If you would like to debug this, the easiest way is to add 
[debug_diff.txt](https://github.com/microsoft/winfile/files/8684330/debug_diff.txt)
and step by step go up and out with Shift F11, and then go down until you arrive around Sel_BlockHiLite().
I am sure with source this more fun:-)

malxau-msft:
_ is an innocent victim.  The issue is that you can't press underscore without holding shift, and holding shift causes the list to multiselect everything between its current location and its new one.  This can be seen by holding shift and pressing any alphabetic character.  This behavior exists in NT 3.1, and the same behavior with _ can be displayed with many other characters - ~, #, ^, +, ...

I don't know if whoever added this to the list box intended it though.  It's interesting that it wasn't carried forward into ListView, so explorer doesn't behave this way.

schinagl:
Then I would go with this
Always select item with WM_CHARTOITEM. See comment in code

src/wfdir.c
```diff
@@ -645,7 +645,15 @@ DirWndProc(
       if (j == cItems)
          return -2L;
 
-      return((i + j) % cItems);
+      int pos = (i + j) % cItems;
```
malxau-msft, line 648: Please avoid C99 variable declarations - they have to be changed to support older systems, and it doesn't seem like we gain anything from using them.

src/wfdir.c
```diff
@@ -645,7 +645,15 @@ DirWndProc(
...
          return -2L;
 
-      return((i + j) % cItems);
+      int pos = (i + j) % cItems;
+
+      // There is a weird behavior in listbox which selects all between anchor an caret
+      // if SHIFT is pressed. Since we return the position here and thus caret will be 
+      // updated, anchor is behind, and pressing shift selects all between anchor and caret
+      // To overcome this we select the current position, and bring anchor and cart in sync.
+      SendMessage(hwndLB, LB_SETSEL, 1, pos);
```
malxau-msft, line 654: (Sorry, this is just a rant.)  LB_SETSEL changes the selection state of an item; in theory, it doesn't move the caret.  However, moving the caret doesn't change the behavior here.  If we did change the selection state ourselves, that suggests we should also return -2, except if we do, the previous selection is retained (ie., `return pos` is used to deselect the previous item.)  It really looks to me like this code works because the listbox has explicit detection for LB_SETSEL which causes it to alter its behavior when this message returns.  Ick.

Also, the tree control duplicates all of this logic, but doesn't have the same problem since it's not multiselect.  Sigh.

malxau-msft (changes requested):

src/wfdir.c
schinagl: I have changed this for here, but I very much prefer C99 declaration since expresses kind of 'locality' much more. Anyhow ;-)

src/wfdir.c
schinagl: This change is about moving the anchor.
The caret is changed due to the return value.
We can not return -2, because we need the new position

What I saw from debugging the problem is that listbox simply selects from anchor to caret. If anchor and caret is the same we are fine. That's what we are doing with is line.

This solution is the simplest way out. It does not cost performance, It simply solves the issue and it documents the workaround.
Everything else means changing code somewhere else.

malxau-msft (approved):

src/wfdir.c
malxau-msft: Right, I see this is the best solution.  What's puzzling me is that this is LB_SETSEL as opposed to LB_SETCARETINDEX or LB_SETANCHORINDEX.  Changing the caret index appears to not change behavior; changing the anchor index means even though it returns pos, the new item is not selected.  So LB_SETSEL is pragmatically the right thing to do, but there's more to it than just moving the anchor.  Thinking about it today, this is probably because Shift+key selects everything from the anchor to the caret, so moving the anchor means it selects nothing, and LB_SETSEL is both moving the anchor and making the selection.  It just seems unexpected (to me at least.)

PR #324, opened 2022-05-13
Do not show the permissions icon in the taskbar on new installations

schinagl:
Do not show the 'permissions' icon in the taskbar with this branch, because it is not available with W7++ anymore.

The coding is still in, because other branches which might target W2K & WXP do have the functionality

### DIFF
--- a/src/tbar.c
+++ b/src/tbar.c
@@ -75,7 +75,8 @@ static TBBUTTON tbButtons[] = {
   {12, IDM_MOVE       , TBSTATE_ENABLED, TBSTYLE_BUTTON, 0 },
   {13, IDM_DELETE     , TBSTATE_ENABLED, TBSTYLE_BUTTON, 0 },
   { 0, 0              , TBSTATE_ENABLED, TBSTYLE_SEP   , 0 },
-  {27, IDM_PERMISSIONS, TBSTATE_ENABLED, TBSTYLE_BUTTON, 0 },
+  // IDM_PERMISSIONS not shown anymore, because there is no way to use the old acledit functionality with W7/10/11
+  // {27, IDM_PERMISSIONS, TBSTATE_ENABLED, TBSTYLE_BUTTON, 0 },
 };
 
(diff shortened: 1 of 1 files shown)


malxau-msft:
Philosophically, I think we should try to allow older systems to use the code in this tree, but that's different to keeping code alive in the tree that can't work on newer systems.

For this specific case though, looking now, I think the "new" APIs are in aclui.dll.  See `EditSecurity` and friends.  There'd still be a lot of wrapper code to interact with them, since WinFile would need to supply `ISecurityInformation`.  I haven't used these but if I'm reading it right, the caller is expected to supply a callback for `OpenElevatedEditor` which could be used if the calling process doesn't have access (so the DLL doesn't directly provide a helper process for this task.)  These "new" APIs appear to support 2000 and up, so keeping code for the older APIs is really for NT 4 and below (which already have an inbox winfile.)

Given that, do you think this is something we should consider integrating with?

schinagl:
My motivation was to short term get rid of a button which can't be used anymore.
Older versions: You anyway have to use your branch, because it has WINVER == 0x500. This branch will anyhow only work with W7++ due to WINVER == 0x600. So .... lets at least remove the button.

I personally do not change permissions that much, and if I do, I always do this via 'Alt-Enter' aka properties, and then follow the standard way. 
Since there is a simple other way to change permissions, I don't see a need in implementing a workaround with ISecurityInformation

schinagl:
Only uncommented the IDM_PERMISSIONS, so that one easily can enable it again, if one day there is functionality available again

malxau-msft:
Like you, I don't use this enough or have enough motivation to tackle this any time soon.  But extending our previous conversations, I think it's possible to build an acledit.dll which exposes a File Manager extension interface, that is built on aclui.dll (ie., just a translation layer.)  That acledit becomes a drop-in that would light up this capability.  It's a little moot unless/until somebody builds one, but it's interesting to note that it should work.

PR #327, opened 2022-05-18
Rename shall select all characters when it is a directory and contains a dot

schinagl:
### Problem
Renaming a directory with . in it assumes the characters behind . are an extension, and does not select them

### How to test
Start   [TestRenameSelectiont.zip](https://github.com/microsoft/winfile/files/8720358/TestRenameSelectiont.zip) from an administrative command prompt in your e.g. temp directory. 

In the directory or file pane navigate to bla.000.000.000 and press F2. All characters are selected now
Do this also for bla.000.000.000 - symlink. Also all characters are selected
When selecting bla.000.000.000.dll, which is a file, only the characters before the extension are selected as in the previous version.

### DIFF
--- a/src/wfdlgs2.c
+++ b/src/wfdlgs2.c
@@ -571,24 +571,37 @@ JAPANEND
       }
 
    case WM_NCACTIVATE:
-      if (IDM_RENAME == dwSuperDlgMode)
-      {
-		size_t ich1, ich2;
-		LPWSTR pchDot;
+      if (IDM_RENAME == dwSuperDlgMode) {
+         size_t ich1, ich2;
+         LPWSTR pchDot;
 
-		GetDlgItemText(hDlg, IDD_TO, szTo, COUNTOF(szTo));
-		ich1 = 0;
-		ich2 = wcslen(szTo);
-		pchDot = wcsrchr(szTo, '.');
-		if (pchDot != NULL)
-			ich2 = pchDot - szTo;
-		if (*szTo == '\"')
-		{
-			ich1 = 1;
-			if (pchDot == NULL)
-				ich2--;
-		}
-		SendDlgItemMessage(hDlg, IDD_TO, EM_SETSEL, ich1, ich2);
+         GetDlgItemText(hDlg, IDD_TO, szTo, COUNTOF(szTo));
+         ich1 = 0;
+         ich2 = wcslen(szTo);
+
+         // Search for extension
+         pchDot = wcsrchr(szTo, '.');
+         if (pchDot != NULL) {
+            TCHAR szTemp[MAXPATHLEN];
+            lstrcpy(szTemp, szTo);
+            QualifyPath(szTemp);
+
+            // Is this a file or directory
+            if (GetFileAttributes(szTemp) & (FILE_ATTRIBUTE_DIRECTORY | FILE_ATTRIBUTE_REPARSE_POINT)) {
+               if (szTo[ich2 - 1] == '\"')
+                  ich2--;
+            }
+            else {
+               ich2 = pchDot - szTo;
+            }
+         }
+         // Make sure we handle " properly with selection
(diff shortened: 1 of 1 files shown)


src/wfdlgs2.c
```diff
@@ -571,24 +569,37 @@ JAPANEND
...
+
+         // Search for extension
+         pchDot = wcsrchr(szTo, '.');
+         if (pchDot != NULL) {
+            TCHAR szTemp[MAXPATHLEN];
+            lstrcpy(szTemp, szTo);
+            QualifyPath(szTemp);
+
+            // Is this a file or directory
+            if (GetFileAttributes(szTemp) & FILE_ATTRIBUTE_DIRECTORY) {
```
malxau-msft, line 588: I know it's unfair to say this in one random case/PR, but the growth of these calls everywhere is bothering me for a couple reasons.

First, and most concretely, the enumeration logic uses Wow64DisableWow64FsRedirection, which these calls don't, so the files referenced here don't necessarily match the ones enumerated.  That could be fixed by just having a wrapper function for GetFileAttributes (and calls to CreateFile, in places like DecodeReparsePoint.)

Second, somewhat more philosophically, is that we already have this information in hand in DTA.  This can be shown by putting a breakpoint on FindFirstFile/GetFileAttributes and switching to details view - all of the file attributes become displayed, without querying the file system.  Obviously the list has to know what's a directory to decide which icons to use.  Calling GetFileAttributes in places like this means network calls while the user is doing some simple UI change, so getting it from DTA seems better.  I know this is a large (and probably unrealistic) thing, but I think we'd be in a better place to add a hash map to the front of DTA and replace these calls with a file name lookup from the context that the window already has.

schinagl: Having a wrapper for GetFileAttributes() which uses Wow64DisableWow64FsRedirection() is a good idea, but I would not like to mix it with this PR. I have created an extra PR for this

About the LpxData cache: 
- In this very PR it is hard to retrieve the cached attributes
- The cache is a left over from Win3.11, when the app could not be sure how fast the OS retrieves data for the file system. You had slow floppies, tapes(?) ....
- If the OS nowadays can not provide the basic attributes of a filesystem fast enough, then this should be documented, and it is a problem of he OS.... which doubt that it is.

malxau: Agree the issues should be separate.

I don't think it's fair to say that performance is an OS problem.  Performance is a function of the speed of the underlying transport, which is outside of the OS control.  The real challenge with any networked resource is that the client is limited in how much it can cache, since state can change by another client.

As far as I can tell:

- Win 3.11 had a synchronous file system API with a cooperative windowing UI.  This means any call to interact with a file prevents UI updates not just from the calling process, but from all processes.  This is why winfile has `wfYield` - that's a 16 bit legacy to let other programs run.
- NT immediately addressed that by ensuring each process is preemptively scheduled, so a slow file system operation only affects UI updates of the calling process.  This still means a process can easily hang itself.  NT provided primitives to prevent this, namely threads, but that requires programs to be restructured to use them.
- The DTA cache was born in NT, specifically in NT 3.5, since that's when it was possible to build.  Unfortunately it was incomplete, since it attempted to asynchronously populate in `CreateDTABlockWorker` but could fall back to synchronous population if it wasn't populated promptly. This makes no sense: if a transport is fast there's no benefit to using a background thread, and if a transport is slow the foreground stalls anyway.  The commit for this change said it was for "ras" (aka dial up) where latency would have been brutal.

In other words, I don't think it's correct to say that synchronous file system calls are the "new" way and the cache is the "old" way; the cache is the new way but it was never finished.  I'll speculate this is one of the reasons explorer didn't leverage much of winfile either - they really wanted to ensure file operations didn't occur on the UI thread, which required a ground-up design.  If they hadn't done this, using explorer on a network resource would prevent the user being able to use the start button, etc.

PR #328, opened 2022-05-21
Hotkey for opening New Window on search result

schinagl:
### Problem

1. There is no hotkey avaiable on the search menu to open the containing folder of an item. This only works via right mouse and select first entry from popup-menu
2. If a search is executed on a hidden or system folder, and one selects to open a 'New Window' from the pop-up on the search outputs, the New Window always shows the root of the active drive.

### Improvements
When selecting an item on the search folder, one can now press CTRL + Shift + Enter to open a 'New Window' with the containing folder from the selected search result.
This now also works if the result was on hidden folder. The New Window opens with the containing folder.

### About the code
The easy part was to extended accelerator table. Fixing this for hidden or system folders is done by forwarding the attributes to the search folder. This was simply missing.


craigwims:
While in the search results page, Alt+W, Alt+N does the same thing -- or am I missing something?

schinagl:
Almost all functionality can be reached via the menu, here ALT+W N. Correct.
I think in this very case it makes sense to use the freely available CTRL + SHIFT + ENTER to open a new window, because it is something you do often.
Furthermore CTRL + SHIFT + Enter this falls into the category of 'doing' something with a file:
ENTER == Execute file
CTRL + ENTER == Exec file as admin
ALT + ENTER == Show Properties of file
CTRL + SHIFT + ENTER == Open a window with the files location

src/lang/res_tr-TR.rc
```diff
@@ -127,7 +127,7 @@ CTXMENU MENU PRELOAD
 BEGIN
     POPUP "Dummy Popup"
     BEGIN
-    MENUITEM    "Seçimden &Yeni Pencere",           IDM_NEWWINDOW
+    MENUITEM    "Seçimden &Ctrl+Shift Enter",           IDM_NEWWINDOW
     MENUITEM    "&Aç\tEnter",                       IDM_OPEN
```
craigwims, line 131: Based on google translate, the phrase Yeni Pencere is the one to keep; not the first word.

src/wfdlgs2.c
```diff
@@ -188,59 +188,62 @@ SearchDlgProc(register HWND hDlg, UINT wMsg, WPARAM wParam, LPARAM lParam)
...
-                     MDICS.style = bMaximized ? WS_MAXIMIZE : WS_MINIMIZE;
-					 MDICS.x  = CW_USEDEFAULT;
-                     MDICS.y  = 0;
-                     MDICS.cx = CW_USEDEFAULT;
-                     MDICS.cy = 0;
+                      SendMessage(hwndMDIClient, WM_MDICREATE, 0L, (LPARAM)(LPMDICREATESTRUCT)&MDICS);
 
-                     SendMessage(hwndMDIClient, WM_MDICREATE, 0L, (LPARAM)(LPMDICREATESTRUCT)&MDICS);
+                      // Forward the attributes to search window
+                      SetWindowLongPtr(hwndSearch, GWL_ATTRIBS, GetWindowLongPtr(hwndMDIChild, GWL_ATTRIBS));
```
craigwims, line 235: Make a note that hwndSearch is created by the above call with WM_MDICREATE.

schinagl: done

Issue #296, opened 2022-01-31
Goto Cache Matching only full words


schinagl:
What is the intended behavior when it comes to matching strings in the Goto dialog:

Assume I have a directory 'c:\bla\ISSUE-123456 TC064\foo.txt'

When I type 'TC064' it matches. So far so good.
If I type '123456' it does not match, but it matches when I type 'ISSUE'. It also does not match when I type 'C064'

I would expect it also to match '123456', or 'C064'

So what should it be by design?

craigwi:
I created this feature and I support improving it.  The data structure is a Trie which is setup to match on word prefixes where words are simply space separated substrings.  A simple change would be to increase the number of punctuation characters.

clzls:
(Have seen it at the bottom of my inbox) The feature seems to work well when a language use spaces to separate words, but it's not the case of Chinese, Japanese or some other language (Vietnamese use spaces to separate syllables!). We do not use space to separate words. So when I try to search "智能"(Intelligence), I would expect "人工智能" (Artificial Intelligence) or "人工智能研究" (Artificial Intelligence Research) in my list.

PR #329, opened 2022-05-25
* Goto Cache Improvements 

schinagl:
### Problems
The Goto Cache had a few little problems
* It followed all reparse points, which could drive it into a reparse - loop. 
* Items were enumerated many times when they were behind reparse points
* The punctuation was limited to \<blank\> See #296
* The punctuation could not be configured

### Changes
* The punctuation is now extended to "- _."
* The punctuation can be configured via Winfile.ini[Settings]GotoCachePunctuation
* The Goto Cache now only follows 
**so called 'outer' reparse points, but drops so called 'inner' reparse points, because these files are anyhow reached via normal file enumeration.
** Onedrive. This PR also works for OneDrive and opposite to #132 has no limitations.

### General
What is a 'inner'/ 'outer' reparse point: 
* With respect to a root, e.g. c:\foo a reparse point is called 'inner' if it points to a location below the root, e.g. c:\foo\bla\reparse -> c:\foo\bla\reparse_dest
* With respect to a root, e.g. c:\foo a reparse point is called 'outer' if it points to a location outside the root, e.g. c:\foo\bla\reparse -> c:\bla\dest

### DIFF
--- a/src/wfgoto.cpp
+++ b/src/wfgoto.cpp
@@ -19,7 +19,7 @@ extern "C"
 #include "lfn.h"
 }
 
-void BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, LPTSTR szRoot, PDNODE pNode);
+void BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, LPTSTR szRoot, PDNODE pNode, DWORD scanEpoc, LPTSTR szCachedRootLower);
 void FreeDirectoryBagOValues(BagOValues<PDNODE> *pbov);
 
 DWORD g_driveScanEpoc;				// incremented when a refresh is requested; old bags are discarded; scans are aborted if epoc changes
@@ -293,7 +293,7 @@ void FreeDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes)
 	delete pbov;
 }
 
-BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes, LPCTSTR szRoot, PDNODE pNodeParent, DWORD scanEpoc)
+BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes, LPCTSTR szRoot, PDNODE pNodeParent, DWORD scanEpoc, LPTSTR szCachedRootLower)
 {
 	LFNDTA lfndta;
 	WCHAR szPath[MAXPATHLEN];
@@ -418,21 +418,25 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes, 
 				// inner peparse points, because they would cause double enumeration
 
 				// If the anchor is not part of the path then it is an outer reparse point
-				// e.g. outer reparsepoint: szCachedRoot == c:\bla, szTemp == d:\foo
-				// e.g. inner reparsepoint: szCachedRoot == c:\bla, szTemp == c:\bla\foo
+				// e.g. outer reparsepoint: szCachedRootLower == c:\bla, szTemp == d:\foo
+				// e.g. inner reparsepoint: szCachedRootLower == c:\bla, szTemp == c:\bla\foo
 				_wcslwr_s(szTemp, MAXPATHLEN);
 				if (!wcsstr(szTemp, szCachedRootLower))
 					bFollow = TRUE;
 				break;
+
+			default:
+				BOOL bFollow = FALSE;
 			}
 		}
-		else
+		else {
 			bFollow = TRUE;
+		}
 
 		if (bFollow)
 		{
 			// add directories in subdir
-			if (!BuildDirectoryBagOValues(pbov, pNodes, szPath, pNodeChild, scanEpoc))
+			if (!BuildDirectoryBagOValues(pbov, pNodes, szPath, pNodeChild, scanEpoc, szCachedRootLower))
 			{
 				WFFindClose(&lfndta);
 				return FALSE;
@@ -718,11 +722,13 @@ BuildDirectoryTreeBagOValues(PVOID pv)
 	WCHAR 	seps[]{ L";" };
 	PWCHAR   token{ nullptr };
 	PWCHAR   szCachedRoot = wcstok_s(szCached, seps, &token);
+	TCHAR    szCachedRootLower[MAXPATHLEN];
+
 	while (szCachedRoot)
 	{
 		lstrcpy(szCachedRootLower, szCachedRoot);
 		_wcslwr_s(szCachedRootLower, MAXPATHLEN - (szCachedRoot - szCached));
-		buildBag |= BuildDirectoryBagOValues(pBagNew, pNodes, szCachedRoot, nullptr, scanEpocNew);
+		buildBag |= BuildDirectoryBagOValues(pBagNew, pNodes, szCachedRoot, nullptr, scanEpocNew, szCachedRootLower);
 
 		szCachedRoot = wcstok_s(NULL, seps, &token);
 	}
--- a/src/wfgoto.cpp
+++ b/src/wfgoto.cpp
@@ -427,6 +427,7 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes, 
 
 			default:
 				bFollow = FALSE;
+				break;
 			}
 		}
 		else {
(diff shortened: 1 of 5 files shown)


schinagl:
Rebased onto master

src/wfinit.c
```diff
@@ -1312,6 +1312,14 @@ JAPANEND
...
    win.rc.bottom -= win.rc.top;
 
+   // We need to know about all reaprse tags
+   hNtdll = LoadLibrary(L"ntdll.dll");
+   if (hNtdll)
+   {
+      pfnRtlSetProcessPlaceholderCompatibilityMode = (RtlSetProcessPlaceholderCompatibilityMode_t)GetProcAddress(hNtdll, "RtlSetProcessPlaceholderCompatibilityMode");
+      if (pfnRtlSetProcessPlaceholderCompatibilityMode)
+         pfnRtlSetProcessPlaceholderCompatibilityMode(PHCM_EXPOSE_PLACEHOLDERS);
+   }
```
craigwims, line 1322: How is the behavior of PHCM_APPLICATION_DEFAULT determined?  Wouldn't it be better to change that?

malxau-msft: (Sorry for being MIA, I've been busy with day job)

I had the same reaction and looked this up the other day.  PHCM_APPLICATION_DEFAULT is zero, which is used because the PEB is initialized with RtlZeroMemory.  The meaning is provided by `HsmOsGetPlaceholderCompatMode`, which is a clean function that's easy to read.  But the high level point is that, if it's not a sync process or an OS process, it's going to get hidden placeholder behavior unless it explicitly asks to observe placeholders.

src/wfinit.c
```diff
@@ -1312,6 +1312,14 @@ JAPANEND
    win.rc.right -= win.rc.left;
    win.rc.bottom -= win.rc.top;
 
+   // We need to know about all reaprse tags
+   hNtdll = LoadLibrary(L"ntdll.dll");
```
malxau-msft, line 1316: This is really minor, but note that ntdll is basically guaranteed to be mapped in any Win32 process.  Using GetModuleHandle allows you to get a handle to the module without a reference (so it doesn't need FreeLibrary) and uses the module name as opposed to file name to find it.  This code is relying on KnownDlls to translate "ntdll.dll" into %SystemRoot%\system32\ntdll.dll to avoid exploits loading from the current directory.

src/wfinit.c
schinagl: Correct. PHCM_APPLICATION_DEFAULT would not give info on reparse points for OneDrive. 
We need this info, so that we can follow reparse points during goto-Cache fill-up

src/wfinit.c
schinagl: Changed it, because it is a few lines code less.
But I guess if a .dll is already mapped a LoadLibray() turns into a GetModuleHandle()
Anyhow it is less code

schinagl:
Used GetModuleFileHandle()

src/wfinit.c
```diff
@@ -1312,6 +1312,14 @@ JAPANEND
    win.rc.right -= win.rc.left;
    win.rc.bottom -= win.rc.top;
 
+   // We need to know about all reaprse tags
+   hNtdll = GetModuleHandle(L"ntdll.dll");
```
craigwims, line 1316: This should use NTDLL_DLL.

src/winfile.h
```diff
@@ -1183,6 +1185,9 @@ Extern TCHAR        szUILanguage[]          EQ( TEXT("UILanguage") );
 Extern TCHAR        szEditorPath[]          EQ( TEXT("EditorPath"));
 Extern TCHAR        szMirrorContent[]       EQ( TEXT("MirrorContent") );
 Extern TCHAR        szCachedPath[]          EQ( TEXT("CachedPath"));
+Extern TCHAR        szCachedRootLower[MAXPATHLEN];
```
craigwims, line 1188: Should be passed as a parameter to BuildDirectoryBagOValues instead of a global.

src/wfgoto.cpp
```diff
@@ -267,13 +267,14 @@ vector<wstring> SplitIntoWords(LPCTSTR szText)
...
-	while (getline(ss, item, L' '))
+	wchar_t tempStr[MAXPATHLEN];
+	wcscpy_s(tempStr, szText);
+	PWCHAR token{ nullptr };
+	PWCHAR word = wcstok_s(tempStr, szPunctuation, &token);
+	while (word)
 	{
-		if (item.size() != 0)
-			words.push_back(item);
+		words.push_back(word);
```
craigwims, line 276: Seems to me we should retain the check on word.size() since it appears wcstok_s will return an empty string if two punctuation are in a row.

src/wfgoto.cpp
```diff
@@ -382,11 +383,60 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes,
...
+				// e.g. outer reparsepoint: szCachedRoot == c:\bla, szTemp == d:\foo
+				// e.g. inner reparsepoint: szCachedRoot == c:\bla, szTemp == c:\bla\foo
+				_wcslwr_s(szTemp, MAXPATHLEN);
+				if (!wcsstr(szTemp, szCachedRootLower))
+					bFollow = TRUE;
+				break;
+			}
+		}
+		else
+			bFollow = TRUE;
```
craigwims, line 430: Given the above (and larger) branch of the if() statement has {}, the else clause should too.

src/wfgoto.cpp
```diff
@@ -382,11 +383,60 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes,
...
+		BOOL bFollow = FALSE;
+		if (lfndta.fd.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT)
 		{
-			WFFindClose(&lfndta);
-			return FALSE;
+			// Check which reparsepoint
+			wchar_t szTemp[MAXPATHLEN];
+			DWORD tag = DecodeReparsePoint(szPath, szTemp, MAXPATHLEN);
+			switch (tag) 
+			{
```
craigwims, line 394: To clarify the cases, I would put a default: bFollow = FALSE;

craigwims:
Nice work; just a small number of changes to be ready to go.

src/wfgoto.cpp
schinagl: wcstok_s() doesn't deliver empty strings. Tried this with e.g. 
`SplitIntoWords(L"_-_test-_-foo_-_");`
It returns 'test' and 'foo'

schinagl:
Code review in-cooperated

src/wfgoto.cpp
```diff
@@ -382,11 +383,64 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes,
...
+				// If the anchor is not part of the path then it is an outer reparse point
+				// e.g. outer reparsepoint: szCachedRootLower == c:\bla, szTemp == d:\foo
+				// e.g. inner reparsepoint: szCachedRootLower == c:\bla, szTemp == c:\bla\foo
+				_wcslwr_s(szTemp, MAXPATHLEN);
+				if (!wcsstr(szTemp, szCachedRootLower))
+					bFollow = TRUE;
+				break;
+
+			default:
+				BOOL bFollow = FALSE;
```
craigwims, line 429: Oops; should not declare a local and be followed by 'break;'

schinagl: My bad. Copy/Paste. Sry

schinagl:
Fixed copy/paste error

src/wfgoto.cpp
```diff
@@ -382,11 +383,64 @@ BOOL BuildDirectoryBagOValues(BagOValues<PDNODE> *pbov, vector<PDNODE> *pNodes,
...
+				// If the anchor is not part of the path then it is an outer reparse point
+				// e.g. outer reparsepoint: szCachedRootLower == c:\bla, szTemp == d:\foo
+				// e.g. inner reparsepoint: szCachedRootLower == c:\bla, szTemp == c:\bla\foo
+				_wcslwr_s(szTemp, MAXPATHLEN);
+				if (!wcsstr(szTemp, szCachedRootLower))
+					bFollow = TRUE;
+				break;
+
+			default:
+				bFollow = FALSE;
```
craigwims, line 429: Since you are very detailed oriented, there should be a break; statement after setting bFollow to FALSE.  Fall through cases, if used, should be explicitly noted in a comment.

schinagl: done

PR #333, opened 2022-06-09
TypeAhead Fix

schinagl:
### General
With #290 TypeAhead was fixed, but we all overlooked one use-case

### Problem

```
as
bla
san
system
syxsser
zzz
```

The [attached .bat file](https://github.com/microsoft/winfile/files/8866981/TypeAHead.zip) creates such a situation.
- Select 'bla' in the right pane
- Type 's' 'y' 's'
- The cursor is on 'syxsser' which is **not correct**. It should be on 'system'

### About the changes
With #290 we overlooked, that only *consecutive same* characters shall move the cursor ahead by one. The #290 solution also moves ahead by one, if any character typed within a word is the same as the first, but other characters are in between.

I now cross checked the fixed behavior with explorer and it is exactly the same.

### DIFF
--- a/src/wfutil.c
+++ b/src/wfutil.c
@@ -1571,6 +1571,7 @@ BOOL TypeAheadString(WCHAR ch, LPWSTR szT)
 {
    static DWORD tick64 = 0;
    static WCHAR rgchTA[MAXPATHLEN] = { '\0' };
+   static BOOL sameChar = TRUE;
    DWORD tickT;
    size_t ich;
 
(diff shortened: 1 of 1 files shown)


malxau-msft (approved):
I tried this out and agree it works as advertised.  That said, I'd like Craig to confirm he's okay with the behavior since he expressed opinions in #290 .  It seems like this is still following the same approach of having two "modes" where typeahead is either looking for the next file starting with the same single char (sameChar == TRUE, return == FALSE) or is looking for a file that contains an entire prefix match (sameChar == FALSE, return == TRUE).

src/wfutil.c
```diff
@@ -1571,6 +1571,7 @@ BOOL TypeAheadString(WCHAR ch, LPWSTR szT)
 {
    static DWORD tick64 = 0;
    static WCHAR rgchTA[MAXPATHLEN] = { '\0' };
+   static BOOL sameChar = TRUE;
    DWORD tickT;
```
craigwims, line 1575: let's add a comment indicating that sameChar means that all of the characters in rgchTA thus far are the same.

src/wfutil.c
```diff
@@ -1586,20 +1587,26 @@ BOOL TypeAheadString(WCHAR ch, LPWSTR szT)
...
+   if (rgchTA[0] == ch && TRUE == sameChar) {
+      // Same consecutive character as the first was pressed so jump ahead by one
       szT[0] = ch;
       szT[1] = '\0';
 
       return FALSE;
    }
+   else
+      // not the same character as the first
+      sameChar = FALSE;
```
craigwims, line 1610: Put {} around this statement because the first part of the if() has them.

craigwims:
The behavior looks good.  A couple of small things to fix.

schinagl:
code review incooperated

PR #334, opened 2022-06-11
Wrapper for GetFileAttributes() with Wow64DisableWow64FsRedirection()

schinagl:
Wow64DisableWow64FsRedirection is not used in all file access
This PR provides a Wrapper WFGetFileAttributes, which does this

### DIFF
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -210,7 +210,7 @@ WFFindClose(LPLFNDTA lpFind)
 BOOL
 WFIsDir(LPTSTR lpDir)
 {
-   DWORD attr = GetFileAttributes(lpDir);
+   DWORD attr = WFGetFileAttributes(lpDir);
 
    if (attr == INVALID_FILE_ATTRIBUTES)
       return FALSE;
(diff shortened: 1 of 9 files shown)


malxau:
Thanks for doing this.  I tried this out, and unfortunately I hadn't thought hard enough about what's happening.  My very first test case was compiling winfile as 32 bit and trying to run `\Windows\System32\OpenSSH\ssh.exe`, which fails in `ShellExecute` saying the file isn't found.  On reflection, making the scenario work implies handling any Win32 path based API - `CreateDirectory`, `CreateFile`, `CreateSymbolicLink`, `CopyFile`, `DeleteFile`, `MoveFile`, all of it.  I'll take a pass at this later today or tomorrow to see what a full change for the scenario looks like.

Alternatively, since this is specific to 32 bit builds, it may be obsolete; but if that's true I think we should remove the call in enumerate, because what happens now is inconsistent (eg. displaying a program and being unable to execute/manipulate it.)

Thanks (and sorry for pointing you on a wild goose chase.)

schinagl:
You were faster than me, because I wanted to ask, how many other calls we have to guard with Wow64Redirection. 

If you ask me, go the safe and simple way, and remove it from enumeration. 

I will abandon this PR and you create one for this removal from enumeration?!

Issue #50, opened 2018-04-10
Path length limit of (I think) 260 characters


PhilippGrasboeckDynatrace:
I really like winfile due to it's responsiveness.
Is there any chance to get rid of the fixed max length of 260 characters for the path length?
If so, then winfile would be truly superior to Windows Explorer, which has the same problem.

daniel-white:
i think many of the win32 api calls have that limit, so its a limitation of win32.

My1:
but there are ways around it. total commander for example can show that easily.

ZanderBrown:
Windows 10 has long path mode

My1:
w10 isnt even needed.

ZanderBrown:
Yes but Total Commander likely isn't using native Win32

My1:
apparently there has already been a not too-bad way before:
https://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx#maxpath

"The Windows API has many functions that also have Unicode versions to permit an extended-length path for a maximum total path length of 32,767 characters. This type of path is composed of components separated by backslashes, each up to the value returned in the lpMaximumComponentLength parameter of the GetVolumeInformation function (this value is commonly 255 characters). To specify an extended-length path, use the "\\?\" prefix. For example, "\\?\D:\very long path"."

malxau:
I think there are three issues here that need to be kept separate.

1. Winfile, like many legacy codebases, uses fixed sized "worst case" string lengths and does things like put them on the stack.  In order to support huge lengths, this type of thinking needs to go away and be replaced by fully dynamic allocations for all file names.  As of this writing, there are 182 instances of "MAXPATHLEN" in *.c files.  Until this is done there's not much point talking about APIs to send huge strings to, because there's no huge strings to send.
2. The Windows 10 API support is opt-in via manifest, but obviously doesn't solve anything pre recent 10.
[...]

PR #340, opened 2022-07-24
Very long path

schinagl:
### Problems
Ever since Winfile supports path lengths up to 256 characters. In certain situations it is useful to have more. Windows supports 32767, but this PR aims less and increases up to 1024
It also addresses the request from #50

### Changes
* MAX_PATH occurrences were replaced with MAXPATHLEN so that there is one common definition for the length
* Added the long longPathAware manifest
* Added verylongpath.reg file which lifts the 256 character limitation for Windows 10 system > 1607. Can be [found here] (https://github.com/microsoft/winfile/files/9175242/TestVeryLongPath.zip)

* Increased the stacksize so that filling the Goto-Cache survives 512 levels of recursion
* Fixed a bug which was there anyhow via  [44991cb]

### How to test
Take [TestDeepPath.bat](https://github.com/microsoft/winfile/files/9175242/TestVeryLongPath.zip) and run it. A path with about 750 chars is created
Now play around, and navigate through it. Do all you think should work

### Status
Basically this PR works quite well, but it is very experimental, because a change like this 
* is hard to test, because all tests are manual
* the achieved test coverage is unknown.
So it is a draft PR, just for all who want to compile their own Winfile and maybe it works

### DIFF
--- a/src/wfcomman.c
+++ b/src/wfcomman.c
@@ -692,8 +692,8 @@ OpenOrEditSelection(HWND hwndActive, BOOL fEdit)
       //
       if (fEdit)
       {
-          TCHAR szEditPath[MAX_PATH];
-          TCHAR szNotepad[MAX_PATH];
+          TCHAR szEditPath[MAXPATHLEN];
+          TCHAR szNotepad[MAXPATHLEN];
 
           // NOTE: assume system directory and "\\notepad.exe" never exceed MAXPATHLEN
           if (GetSystemDirectory(szNotepad, MAXPATHLEN) != 0)
@@ -701,7 +701,7 @@ OpenOrEditSelection(HWND hwndActive, BOOL fEdit)
           else
               lstrcpy(szNotepad, TEXT("notepad.exe"));
 
-          GetPrivateProfileString(szSettings, szEditorPath, szNotepad, szEditPath, MAX_PATH, szTheINIFile);
+          GetPrivateProfileString(szSettings, szEditorPath, szNotepad, szEditPath, MAXPATHLEN, szTheINIFile);
 
           CheckEsc(szPath);     // add quotes if necessary; reserved space for them above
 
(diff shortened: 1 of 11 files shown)


schinagl:
Had Winfile.exe with 1024 characters running for a week during my everyday use, and it worked flawlessly. 
There seem to be no sideeffects on the operations <= 256 characters and the operations > 256 also work really well. Who ever wrote this thing 30 years ago and implemented the limits in WinFile, really did it very well :-)

Anyhow it would be nice if you could help testing this. A version from my [private](https://github.com/schinagl/winfile/tree/_hermann) branch
[can be downloaded here](https://schinagl.priv.at/nt/winfile/winfile.zip). I called this version 10.1.4.42. This is just **my** internal number.

Most likley there will be an official version with precompiled binaries having many changes from the last year in during the course of this year.

Thx for your help in testing this.

schinagl:
Rebased onto #341 merge to master

src/wfcomman.c
```diff
@@ -601,7 +601,7 @@ OpenOrEditSelection(HWND hwndActive, BOOL fEdit)
    DWORD ret;
    HCURSOR hCursor;
 
-   WCHAR szPath[MAXPATHLEN+2];  // +2 for quotes if needed
+   WCHAR szPath[MAXPATHLEN * 2 + 2];  // +2 for quotes if needed, * 2 because CreateDirWindow() assumes a filename can be the same length as MAXPATHLEN
```
craigwims, line 605: Let's separate out this change per #344 .

schinagl:
Resolved conflicts solution brought in with #344
Basically it now comes down to a few very simple things
- Replace MAX_PATH with MAXPATHLEN in a few places, because MAXPATHLEN is the winfile-way to express this.
- Added the longPathAware manifest to the .manifest file
- Increased the stack size in the .vcxproj so that filling the Goto-Cache survives 512 levels of recursion. (tried it)

Voila, and this works perfect since months on my site

schinagl:
Any news here? Can we add this since it almost contains no code changes anymore.... ?

craigwims:
reviewing again.

craigwims:
I'm good with this PR.  Shall I merge it now or wait for #312?

schinagl:
Merge it. Will rebase 312 if needed

Issue #338, opened 2022-06-17
The folder contents jumping to the top when you open it.  Sigh... 


5YNTH3T1K:
I want an alternative to window Files explorer for this one single reason. 

The other reason is I just want to look at my PC as a Drive and not a whole bunch of virtual nonsense, 
"This PC, desktop, documents etc. All complete and utter MS nonsense. 

So it's really simple: 
If I open the "Windows" folder, for example, the contents listing jumps to the top of the pane. 
NO !!!!! this is REALLY BAD. It's been really bad since XP and I am totally sick of it. 
Things that jump around the screen are a terrible design "feature". 

The folder should open and that root folder should stay under the mouse. The folders inside that folder should 
open below that. Just like XP did. and No other file explorer known to man does this anymore. 

Can File Commander over come this problem OR is it baked into windows?  

I am super interested to get rid of this feature. Will pay $$$ !!! I am serious. 
I cannot stand it and when open larger volume folders it drives me crazy. 

Can you help ? 

Thanks! 

![image](https://user-images.githubusercontent.com/46511734/174395143-4d36b74b-5c7e-4a83-833f-cf7050fb12be.png)

![image](https://user-images.githubusercontent.com/46511734/174395173-65213a2a-9ec9-45ce-857c-332cb89f317f.png)

craigwims:
I agree that is annoying.

5YNTH3T1K:
> I agree that is annoying.

I have tried other alternatives to windows file explorer and they all do it. Soooo... I am guess that it's part of the windows Visual basic ... which is what everyone seems to use to make software for windows.... 

If you look at the two screen grabs you can see that in the top one there is no "scroll bar" and int eh bottome one there is a "scroll bar" AND it's scrolled down enough to make the contents of the folder at the top. IF the scroll bar was at the top then everything would be fine. Some one decided that this was OK back at MS and they need to learn a thing or two about managing things... such as things jumping around is very very very bad... 

: - )

schinagl:
I doubt this is baked into Windows. I assume there is code somewhere in winfile which does the mentioned behavior.

5YNTH3T1K:
Any ideas on how to cure it? 
It's the worst thing ever when trying to manage loads and loads of folders etc. 
Never did this on Windows XP.  etc.  
It's driving me crazy. 
If I had hundreds of folders to sort through I would refuse point blank to use windows.

craigwi:
The behavior is due to:

[...]

PR #341, opened 2022-07-27
Control tree view scroll behavior on expand via winfile.ini[Settings]…

schinagl:
Control tree view scroll behavior on expand via winfile.ini[Settings]ScrollOnExpand. Default == TRUE
Solves #338


5YNTH3T1K:
This is awesome! Thank you!

5YNTH3T1K:
This is very special. this single problem has vexed people since, um, post Win XP.  All users of MS windows have this same problem and it's very very bad if you have a lot of folders to manage. MS should pay attention to this! Salute!

PR #352, opened 2022-10-29
Symbolic Link, Hardlink and Junction creation is available via menu

schinagl:
As mentioned this is the small part, which provides menu entries for symbolic/hardlink/junction creation


src/lang/res_de-DE.rc
```diff
@@ -244,6 +248,7 @@ BEGIN
     IDS_RENAME,             "Umbenennen"                        /* 32 */
     IDS_FORMAT,             "Datenträger formatieren"                   /* 32 */
     IDS_FORMATSELDISK,      "Wählen Sie den Datenträger aus, den Sie formatieren möchten."                                                                                               /* 128 */
+    IDS_SYMLINK, 			"Symlinken"                       /* 32 */
```
craigwims, line 252: IDS_HARDLINK is missing.

src/res.rc
```diff
@@ -48,6 +48,8 @@ BEGIN
     VK_F7,      IDM_MOVE,       VIRTKEY
     VK_F8,      IDM_COPY,       VIRTKEY
     VK_F9,      IDM_COPYTOCLIPBOARD, VIRTKEY
+    VK_F10,     IDM_SYMLINK,    VIRTKEY
+    VK_F10,     IDM_HARDLINK,    VIRTKEY, SHIFT
     VK_DELETE,  IDM_DELETE,     VIRTKEY
```
craigwims, line 53: F10 is used for the context menu key; what about F11?

src/wfcopy.c
```diff
@@ -2339,8 +2339,9 @@ WFMoveCopyDriverThread(LPVOID lpParameter)
          // deal with case where directory is implicit in source
          // move/copy: *.* -> c:\windows, c:\windows -> c:\temp
          // or foo.bar -> c:\temp
+         // do this not for jucntion or symlinks
 
-         if (!IsWild(pCopyInfo->pTo) && (ManySource || IsDirectory(pCopyInfo->pTo))) {
+         if (!IsWild(pCopyInfo->pTo) && (ManySource || IsDirectory(pCopyInfo->pTo)) && pCopyInfo->dwFunc != FUNC_LINK && pCopyInfo->dwFunc != FUNC_JUNC) {
             AddBackslash(pCopyInfo->pTo);
```
craigwims, line 2345: Based on FUNC_JUNC and FUNC_HARD both go to WFJunction(), why do this for one and not the other?

src/wfdlgs2.c
```diff
@@ -729,16 +766,39 @@ JAPANEND
...
+            case IDM_SYMLINK:
+               pCopyInfo->dwFunc = FUNC_LINK;
+               break;
+            case IDM_HARDLINK:
+               // If source is a directory then create a junction
+               if (IsDirectory(pCopyInfo->pFrom))
+                  pCopyInfo->dwFunc = FUNC_JUNC;
+               else
+                  pCopyInfo->dwFunc = FUNC_HARD;
+               break;
```
craigwims, line 792: In the drag drop code the three cases are distinguished (Symbol, Hard, Junction).  Here the choice between the later two is determined by whether the source is a directory.  Why the difference?  Should we do the same in the drag drop code?

src/lang/res_de-DE.rc
schinagl: Fixed with 9d031f2

src/res.rc
schinagl: Good point!
Fixed with 9d031f2

src/wfcopy.c
schinagl: The above check is either for FUNC_JUNC or FUNC_LINK
We need this here because the 'drap/drop coding' provides pCopyInfo->pTo already with appended *.*, but the 'menu coding' does not
So we must filter it and append *.* so that the later hydraulics work.

src/wfdlgs2.c
schinagl: The 'drag/drop code' already provides all info needed:
- operation type (FUNC_xxxx)
- object type (file or dir)

The Hardlink menu only provides
- object type (file or dir)

To properly perform the operation in WFMoveCopyDriverThread() we need both 'operation type' and 'object type'.
That's why we deduce the operation type here by determining if it is a dir or not, because junctions can only be created from dirs and hardlinks can only be created files.

In other words: 'drag/drop code' provides everything needed, the 'menu code' not.

One other way would have been ( I tinkered about that a long time )
- Have an extra menu for Junctions, so that we end up here with e.g. IDM_JUNCTON. Then we would have the info needed for WFMoveCopyDriverThread() without 'if'

But I didn't want to clutter the menu with an extra entry for Junction, so i 'hid' junction creation behind the 'Hardlink' menu entry.

Well we could improve the situation by having an improved text on the menu: 'Hardlink/Junction\tShift+F11'
But I decided not to do because this also raises the question with the user: 'Why is Hardlink and Junction almost the same?'

So I decided to go the easy way and 'forgot' the 'Junction' on the menu.

But yes, you spotted a grey area and there are arguments for one side or the other.

src/wfcopy.c
schinagl: The above check is either for FUNC_JUNC or FUNC_LINK, which go different ways in the end.

- The 'menu code' already provides the fullly assembled destination for the symlink/junction. One must not append a \*.\*
- The 'drag/drop code' provides the path where the destination should go in and has \*.\* appended (which is later stripped off)

So this code is here to have WFMoveCopyDriverThread() bullet proof for input like that.

src/lang/res_zh-CN.rc
```diff
@@ -245,8 +249,8 @@ BEGIN
     IDS_RENAME,             "重命名"                        /* 32 */
     IDS_FORMAT,             "格式化磁盘"                   /* 32 */
     IDS_FORMATSELDISK,      "选择要格式化的磁盘。"
-
-//    IDS_MAKESYSDISK,      "Make System Disk"              /* 32 */
+    IDS_SYMLINK,            "符號鏈接"                       /* 32 */
+    IDS_HARDLINK,           "硬鏈接"                      /* 32 */
```
clzls, line 253: These are `zh-Hant` variation of Chinese, not `zh-CN`.

src/lang/res_en-US.rc
```diff
@@ -133,6 +135,8 @@ BEGIN
     MENUITEM    "Edit\tF12",	    IDM_EDIT
     MENUITEM    "&Move...\tF7",     IDM_MOVE
     MENUITEM    "&Copy...\tF8",     IDM_COPY
+    MENUITEM    "S&ymlink...\t10", IDM_SYMLINK
```
clzls, line 138: `\t10`? Isn't it `\tF11` or something? 😕

src/lang/res_zh-CN.rc
clzls: `符號鏈接`→`符号链接`，`硬鏈接`→`硬链接`

src/wfdlgs2.c
```diff
@@ -487,6 +487,40 @@ JAPANEND
...
+
+            p = GetSelection(0, NULL);
+
+            LoadString(hAppInstance, IDS_HARDLINK, szTitle, COUNTOF(szTitle));
+            SetWindowText(hDlg, szTitle);
+
+            if (bJAPAN) {
+
+               //by yutakas 1992/10/29
+               LoadString(hAppInstance, IDS_KK_COPYFROMSTR, szStr, COUNTOF(szStr));
```
clzls, line 501: Japanese resource string `IDS_KK_COPYFROMSTR` means `Copy source` so maybe not so suitable here. This if block seems to be some obsolete code...
Maybe @craigwi knows about the history of JAPANBEGIN and JAPANEND?

tokomio-sun:
@clzls 
@craigwims 

thank you for reviewing.
I fixed full width '％' in Line 420.

> Edit: and verbs are using different voices across these messages... Maybe 選択している or such is better?

I should fix.

FROM:
'ファイルを %s にリンクする'

TO:
'%s にファイルをリンク中'

src/lang/res_zh-CN.rc
schinagl: Thx! Resolved with 7dbc75d

src/wfdlgs2.c
schinagl: I know. 
But since all language .rc files contained IDS_KK_COPYFROMSTR I decided to copy the passage in the code.
It looks like as if this means 'From', 'To' and then it is harmless, because it fits for symlink/hardlink creation

craigwims:
No, I don't have any insights into JAPANBEGIN and JAPANEND.

@schinagl, related to an earlier question about the difference between the control afforded by the menus and that of the drag/drop: I was not thinking of adding another menu item, but would like to consider restricting the drag/drop case to match the menu case.  That is, use the logic on line 788 of wfdlgs2.c in the drag/drop code, which would simplify the cases.

src/wfdlgs2.c
clzls: > I know. But since all language .rc files contained IDS_KK_COPYFROMSTR I decided to copy the passage in the code. It looks like as if this means 'From', 'To' and then it is harmless, because it fits for symlink/hardlink creation

The problem is it means "**copy** from" / "Specify a **copy** source" rather then just `From`... may be some kind of confusing.

schinagl:
@craigwims please give a code snippet where you would like to place 
`if (IsDirectory(pCopyInfo->pFrom))`
otherwise we run in danger to talk about different things

As mentioned above the idea in general is:
WFMoveCopyDriverThread() gets PCOPYINFO *fully* filled out from 
- 'drag and drop' coding in wfcopy.c:3519
- 'menu coding' in wfdlgs2.c: 802

As of now WFMoveCopyDriverThread()  is prepared to handle a fully filed out PCOPYINFO properly for the new FUNC_xxxx

The 'menu coding' of this PR now also needs to out fill PCOPYINFO properly in wfdlgs2.c:788

Anyhow please give a code snippet on what you would like to change in detail, so I can give a better answer.

src/wfdlgs2.c
schinagl: I understand the coding now. It seems to handle a Japanese specialty:

All languages basically have 'From:', 'To:' (without verb) on this dialog, but Japanese has 'Copy from'<in Japanese> and 'Copy To' <in Japanese> on the dialog (with a verb).

@tokomio-sun : Do we really need a verb? Could we not just only have 'From:', 'To:' without verb in Japanese?

As of this PR I would keep it as it is, even if it says 'Copy from' when it is 'Hardlink from'.
We can cleanup this up in an extra PR.

craigwims:
Line 316 of wfdirsrc.c: instead of setting iShowBitmap to DROP_HARD unconditionally, set to DROP_JUNC when the source of the drag/drop is a directory.  The case in treectl.c, line 2903 always uses DROP_JUNC since the source there is always a directory.

src/wfdlgs2.c
clzls: > I understand the coding now. It seems to handle a Japanese specialty:
> 
> All languages basically have 'From:', 'To:' (without verb) on this dialog, but Japanese has 'Copy from' and 'Copy To' on the dialog (with a verb).
> 

I see in `winfile_ja-JP.dlg` that `MOVECOPYDLG`'s `From` is translated as `Move from`, so they do need some code to change the verbs when the dialog is reused for other operations. (...and "removing them" seems to be not so "harmless" ... my mistake)

Although I have no evidence, but I guess `bJAPAN` was something served as a localization method long long ago and may be deprecated nowadays by using multiple `.dlg` and `.rc`s, and I didn't see any code that set `bJAPAN` to `true` after it is given `false` in `winfile.h`... Maybe in another PR we could try to fix this with native developers...

schinagl:
I guess I got your point
### What I understood
Your concern starts in wfcopy.c:2776 and wfcopy:2810, because FUNC_HARD is used with oper == OPER_MKDIR
In prosa: 
- in this part a junction comes in and disguises as FUNC_HARD.
- it comes from wfdirsrc.c:316 because a dir is dragged in the file pane and thus is handled via FUNC_HARD

With the proposed change in wfdirsrc.c:316 to send FUNC_HARD or FUNC_JUNC according to the output of IsDirectory() the lines wfcopy.c:2776 and wfcopy:2810 could be deleted and thus would be more logical to an future developer.

### Discussion
- We must not call IsDirectory() in wfdirsrc.c:316 because this code runs every time the mouse moves a pixel. Having a filesystem access here will destroy any performance.
- As far as I know we do not have any other inidcator of attributes here

So we must do this in wfcopy:3495. But here things also will not work because
- directories come in properly and pFrom can be checked
- files come in here quoted and can be many. e.g. "\"bla - Copy - Hardlink.txt\" \"bla - Copy.txt\"". So one has to 
-- parse them into single items
-- unquote them and check them

So 
- unquoting and handling many is something which WFMoveCopyDriverThread() already does.
- should we do all in wfcopy:3495 beforehand? No, this is double effort.
- and whatbout having different types if you have many. e.g there is a file and a directory selected? Which FUNC_xxxx would you assign pCopyInfo->dwFunc?

### Conclusio
Give it a try in the compiler yourself, you will see this would not work. Or create lots of unneccessary effort.
We have to live with this disguise of junctions as FUNC_HARD in wfcopy.c:2776 when they are dragged from the file-pane. 
But we should comment it.

craigwims:
Doing a file system operation on each mouse move is not a good idea; agreed.

Looking at this further, and reflecting on your comments, the difference between *_HARD and *_JUNC seems very, very small.  Why do we have both since at the bottom they both call WFJunction() which does not take FUNC_* as a parameter?

src/wfdlgs2.c
schinagl: Correct it is all about the verb and re-using one dialog for many purposes/verbs
Currently it is up to @tokomio-sun  or @Takym tell us about the necesseity of verbs in this kind of dialog

schinagl:
In most places we need a distinction 
- wfcopy.c:1812 and wfcopy.c:2475 need it when we create -Copy/-Symlink/- Junctions/- Hardlink in the same directory. See #313
- Furthermore the hydraulics of printing out texts on dialogboxes also needs a differentiation. See lang\res_en-US.rc:303. This is a tricky index driven way of finding the proper strings (Was there ever since, but I extended it in the spirit of the author of this part)

In general I also wanted to separate this for readability and debug-ability. The only place where it didn't work out, was wfcopy.c:2776 where a Junction disguises as FUNC_HARD, when it comes from the file-pane

src/wfdlgs2.c
Takym: Yes, these need verbs because "元" and "先" are suffixes.
I think "From"/"To" should be translated to "リンク元"/"リンク先" ("リンク" means "link") in `IDM_HARDLINK` and `IDM_SYMLINK`.

schinagl: @Takym : Can you please provide me 4 strings

'hardlink from'
'hardlink to'
'symlink from'
'symlink to'

I will add respective IDS_KK_xxxxSTR to the resources and make the dialogs also work properly beneath if (bJapan)

Takym: @schinagl I translated all below:

'hardlink from' -> 'ハードリンク元'
'hardlink to' -> 'ハードリンク先'
'symlink from' -> 'シンボリックリンク元'
'symlink to' -> 'シンボリックリンク先'

schinagl: Thx! Added with b678626

craigwims:
A refinement of my idea above: use the logic at line 788 in wfdlgs2.c at line 3506 in wfcopy.c.  In this case, we don't need DROP_HARD (always use DROP_JUNC), but we do need FUNC_HARD .vs. FUNC_JUNC.

schinagl:
### Discussion
Please read [my comment from above](https://github.com/microsoft/winfile/pull/352#issuecomment-1311482881), I discussed this way already:

This does not work because (the most important of many reasons)
- pCopyInfo->pFrom can contain **many** items in a quoted form ""MyFile.txt" "MyDirectory""
- So you can't put pCopyInfo->pFrom into IsDirectory() out of the box 
-- You have to parse pCopyInfo->pFrom into single items and unquote it.
-- Assuming you have unparsed this here: On which of the many items are you going to use IsDirectory(), if you have e.g. a selection containing a file and a directory as outlined in my example above?
- Which FUNC_xxxx would you assign pCopyInfo->dwFunc?. PCOPYINFO may contain **many** files but **one** operation

Please try this in the compiler yourself!

There are many other reasons from [here](https://github.com/microsoft/winfile/pull/352#issuecomment-1311942454) and additionally: 
- lang\res_en-US.rc:303. How would you create different error messages with a mixed type FUNC_HARDJUNC? wfcopy:3301 passes this on. Have a 'if IsDirectory()' here too? You would break the idea of how they selected texts.
- wfcopy:2344 You need to know if it is a directory like type by checking FUNC_JUNC. With a mixed type FUNC_HARDJUNC you don't know this


### Summary
- If it is a 5 line change without any risk no problem, lets do it
- But it is **not a small** change to get rid of one single #define.

schinagl:
But maybe there is a more radical way .... Trying out things ....

schinagl:
The point of my above argument was IsDirectory() on **many** This would not work

But more radical .... If we 
- completely get rid of DROP_JUNC 
- almost get rid of FUNC_JUNC, except for wfcopy:3301, where make a hack for the error message
then things work.

src/wfcopy.c
```diff
@@ -3295,8 +3296,10 @@ WFMoveCopyDriverThread(LPVOID lpParameter)
             }
          }
 
-         ret = CopyError(szSource, szDest, ret, pCopyInfo->dwFunc,
-            oper, bErrorOnDest, bFatalError);
+         INT errorIndex = FUNC_HARD;
+         if (pCopyInfo->dwFunc == FUNC_HARD && IsDirectory(szSource))
+           pCopyInfo->dwFunc = FUNC_JUNC;
```
craigwims, line 3301: I think you meant to set errorIndex here, not change pCopyInfo->dwFunc.

PR #354, opened 2022-10-31
Search window immediatley opens after begin of search

schinagl:
Search window immediatley opens after begin of search

### DIFF
--- a/src/wfdlgs2.c
+++ b/src/wfdlgs2.c
@@ -106,170 +106,145 @@ INT_PTR
 CALLBACK
 SearchDlgProc(register HWND hDlg, UINT wMsg, WPARAM wParam, LPARAM lParam)
 {
-  LPTSTR     p;
-  MDICREATESTRUCT   MDICS;
-  TCHAR szStart[MAXFILENAMELEN];
+   LPTSTR     p;
+   MDICREATESTRUCT   MDICS;
+   TCHAR szStart[MAXFILENAMELEN];
 
-  UNREFERENCED_PARAMETER(lParam);
+   UNREFERENCED_PARAMETER(lParam);
 
-  switch (wMsg)
-    {
-      case WM_INITDIALOG:
+   switch (wMsg)
+   {
+   case WM_INITDIALOG:
 
-          SendDlgItemMessage(hDlg, IDD_DIR, EM_LIMITTEXT, COUNTOF(SearchInfo.szSearch)-1, 0L);
-          SendDlgItemMessage(hDlg, IDD_NAME, EM_LIMITTEXT, COUNTOF(szStart)-1, 0L);
+      SendDlgItemMessage(hDlg, IDD_DIR, EM_LIMITTEXT, COUNTOF(SearchInfo.szSearch) - 1, 0L);
+      SendDlgItemMessage(hDlg, IDD_NAME, EM_LIMITTEXT, COUNTOF(szStart) - 1, 0L);
 
-          GetSelectedDirectory(0, SearchInfo.szSearch);
-          SetDlgItemText(hDlg, IDD_DIR, SearchInfo.szSearch);
+      GetSelectedDirectory(0, SearchInfo.szSearch);
+      SetDlgItemText(hDlg, IDD_DIR, SearchInfo.szSearch);
 
-          p = GetSelection(1, NULL);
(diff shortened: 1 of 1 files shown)


src/wfdlgs2.c
```diff
@@ -243,6 +243,13 @@ SearchDlgProc(register HWND hDlg, UINT wMsg, WPARAM wParam, LPARAM lParam)
...
                           SendMessage(hwndMDIClient, WM_SETREDRAW, TRUE, 0);
                           RedrawWindow(hwndMDIClient, NULL, NULL, RDW_INVALIDATE | RDW_ALLCHILDREN);
+                      } else {
+                        // Show search window immediatley
+                        SendMessage(hwndSearch, FS_CHANGEDISPLAY, CD_PATH, (LPARAM)SearchInfo.szSearch);
+                        ShowWindow(hwndSearch,
+                          GetWindowLongPtr(hwndMDIChild, GWL_STYLE) & WS_MAXIMIZE ?
+                          SW_SHOWMAXIMIZED :
+                          SW_SHOWNORMAL);
                       }
```
craigwims, line 253: This does not appear to cover all of the cases.  Based on some trials, it appears that we need to move the ShowWindow outside of the else clause and just before the break.  Then get bMaximized before the test for hwndSearch.

schinagl: 5553c80 was my best guess. I detail I didn't get 'Then get bMaximized before the test for hwndSearch'.

I moved ShowWindow() out one else clause, but not direct before the break; because bMaximised would be out of scope otherwise.

I tried and played around but the dearch window always opened.
What was the use-case which caused the problems on your side?

craigwims: If the search window is open, but not the top window is the case.  In my test case, I moved the retrieval of bMaximised above if (hwndSearch) and then you can use it just before the break.

schinagl: 529726c should do it.

src/wfdlgs2.c
```diff
@@ -234,17 +236,25 @@ SearchDlgProc(register HWND hDlg, UINT wMsg, WPARAM wParam, LPARAM lParam)
...
                           RedrawWindow(hwndMDIClient, NULL, NULL, RDW_INVALIDATE | RDW_ALLCHILDREN);
+                      } else {
+                         SendMessage(hwndSearch, FS_CHANGEDISPLAY, CD_PATH, (LPARAM)SearchInfo.szSearch);
                       }
                   }
+
+                  // Show search window immediatley
+                  ShowWindow(hwndSearch,
+                     GetWindowLongPtr(hwndMDIChild, GWL_STYLE) & WS_MAXIMIZE ?
+                     SW_SHOWMAXIMIZED :
```
craigwims, line 255: Seems to me this should use bMaximized, not fetch the style of the child.

schinagl: Changed with 503da52.
Sry for the confusion, but up till now I was not able to reproduce the bad-case, So I was in the dark.
Anyhow, I hope it is fine now and solves the 'bad use-case'.

schinagl: @craigwims: Is this one good to go?

craigwims:
There are a set of cases to handle.  Search window: open or not, maximized or not, and I think search underway or not.  When I looked at this last, the PR didn't handle all of the cases.

schinagl:
Added one more line to bring window to the TOP most z-order

### When brought in the situation below and one presses Press CTRL-F and search ==> Search window opens
- Search window not open, other windows normal size, 
- Search window not open, all windows maximised

- Search window open, normal size, searchwindow in front
- Search window open, normal size, searchwindow in the back aka not visible
- Search window open, all windows maximised, searchwindow in the back aka not visible

- Search window open, search window minimized

### Other Use Cases
- Search window open, search running. press hide. if you press CTRL-F again, the Hide/Cancel menu comes as usual

Did I miss a case?

craigwims:
When reviewing this again, I noticed that the old logic fiddled with the redraw of the mdi client because they didn't want the window to be visible at first.  I tried the following simpler version and it seems to work well: https://gist.github.com/craigwims/29c512da0d46d51fb280e9e4fff4c7a2.

schinagl:
Took over Craigs improved version

Issue #356, opened 2022-11-10
New Release in the works


craigwims:
Hi Everyone, I am working on a new release through all of the current channels.  The Microsoft internal process of signing binaries such as winfile.exe has changed.  Once I have that figured out, I will make the release.

In the meantime, I request that folks who can pull the latest sources build and test the changes that have been added over the last many months.  Do report any new issues!

Craig.

craigwims:
The internal tools require VS 2019 tools or later.  We are currently on 2017.  Do folks have a preference about using 2019 .vs. 2022?

Also, the tools require Control Flow Guard (/guard:cf).  @schinagl and others, any issues with that?

The tooling changes I'm testing are currently staged in the branch craigwi-use-2019-toolset-142 which can only be tested by running through the internal ADO instance.

schinagl:
> The internal tools require VS 2019 tools or later. We are currently on 2017. Do folks have a preference about using 2019 .vs. 2022?
> 
> Also, the tools require Control Flow Guard (/guard:cf). @schinagl and others, any issues with that?

I have VS2019 and VS2022, thus I am fine.
CFG can be enabled since VS2015, so ok too

malxau:
> The internal tools require VS 2019 tools or later. We are currently on 2017. Do folks have a preference about using 2019 .vs. 2022?

No...although I do have a preference for setting WindowsTargetPlatformVersion to 10.0, which wasn't possible on 2017.   10.0 means "any 10.0 SDK" as opposed to mandating a specific SDK build.  This is very useful since there's now 12 of them, and having to download/keep a few GBs of space for every project requesting a specific one gets annoying.  There's also a WindowsTargetPlatformMinVersion in case some minimum version is required, where a newer one will suffice.

(Of course, not specifying this also means that different developers may have different header versions.  I doubt that's going to be a big deal because winfile doesn't generally need bleeding edge changes.)

schinagl:
Do you have a rough schedule on the release?
Will it be a x-mas or a 2023 spring release?
Which PRs do we take?

I would like to have #352, #354 and #359 in if possible.

@malxau : which of your PRs should go in?

craigwims:
I got the internal signing process to work before I had travel for thanksgiving.  I should be able to finish before mid-December.  I would like to take the PR that are mentioned by @schinagl above. 

Most important to me is for testing of features already committed.

schinagl:
Something else came to my mind:
[...]

PR #362, opened 2022-12-05
Craigwi use 2019 toolset

craigwims:
Prepare for release using internal Microsoft built pipeline, which requires 2019 toolset and some specific settings.

@schinagl, can you review?

### DIFF
--- a/samples/addon/AddonSample.vcxproj
+++ b/samples/addon/AddonSample.vcxproj
@@ -30,26 +30,26 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <WholeProgramOptimization>true</WholeProgramOptimization>
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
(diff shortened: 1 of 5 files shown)


schinagl:
I compiled with recent VS2019 build tools and SDK 18362. 
Did a quick smoketest in rel/dbg x86/x64 ==> Fine

During compilation and checking the diffs a few things came to my mind

- VerifyResources does not compile in x64
- The version number is still 10.1.4.0. Should it be already 10.2.0.0?
- Debug|Win32 has no /guard:cf : ´<AdditionalOptions>/stack:15000000 /guard:cf %(AdditionalOptions)</AdditionalOptions>`. ok?
- Release|Win32 has no /guard:cf : ´<AdditionalOptions>/stack:15000000 /guard:cf %(AdditionalOptions)</AdditionalOptions>`. ok?
- We dont't need the ControlViewguard on Exeviewer, VerifyResources, FileSignatureInfo, AddOnSample to make it through the build pipe?

Little limitations of my review
- I can't run it on a ARM Hardware, but I suppose this will work. @all: Is there someone who can give it a smoke test on ARM?

craigwims:
Thanks @schinagl.  VerifyResources doesn't need to compile x64 and I won't be building any of the extra tools in the offical build pipeline.  I changed the version number as you saw.  

Due to the lack of internal documentation and there being no local build to verify the settings I had to try different combinations until I got success.  Win32 uses SEH; x64 uses Control Guard.  I could probably remove the `<ControlFlowGuard>Guard</ControlFlowGuard>` in ClCompile for Win32.  

I will confirm the above and create builds to verify.

schinagl:
Perfect☺️

malxau:
```
Is there someone who can give it a smoke test on ARM?
```

I guess that's me.

Firstly, note that the "ARM" architecture is basically legacy as far as I can tell.  Any new ARM device today is ARM64; ARM was for Windows RT, which doesn't officially support Win32 applications.  As in, AFAIK, the set of devices that can run WinFile ARM but not ARM64 is zero.

The result of the smoke test is that the binary doesn't run because it's missing vcruntime140.dll et al, the ARM(32) versions of those DLLs are not included in the ARM64 redistributable package, and there's no public ARM32 package (see http://aka.ms/vcredist ).  They are included with the compiler, so it's possible to copy files around and execute, but this seems like it will cause problems for users.

I pushed a change that renames ARM to ARM64, and I think we should do something along those lines: https://github.com/malxau/winfile/tree/arm-to-arm64

Some notes:
- This change also sets the SDK to 10.0 (so it can build on my system with 19041 but not 18362.)  That's a separate change so I'd understand if people don't want to take it, but it sure makes my life easier.
- I don't have the infrastructure for Store packaging, so the Package.wapproj changes are not tested.  See in particular the change to AppxBundlePlatforms.
- Some of the tools weren't previously building as ARM32, and this change doesn't add ARM64 for those (although it should.)  Also, as you two hit, VerifyResources doesn't compile as 64 bit and doesn't compile as ARM64 for the same reasons.
- In general ARM64 should have the same configuration as x64.  The x64-on-ARM64 layer is implemented using a single set of DLLs with import trampolines, which requires the API surface to be almost identical between the two.
- This gets it out of the business of WOW64 redirection, since ARM64 is the native system architecture.

craigwims:
yes, let's move to ARM64 without the SDK change (for now).

craigwims:
The builds work as noted in the upcoming release issue #356.

PR #363, opened 2022-12-06
Arm to arm64

malxau:
This is just the ARM(32) to ARM64 rename with the SDK change removed.  (That means I can't actually test compile it, but in theory it's a more narrow change.)


craigwims:
@malxau, can you rebase on to master?  I'd prefer to have this show as a separate commit.  Thanks!

PR #364, opened 2022-12-07
Move the ARM build to ARM64

malxau:
This removes the existing 32 bit ARM build support and replaces it with 64 bit ARM build support.  It's largely a rename except for some minor settings.


craigwims:
Local build and internal Microsoft build pipeline works.

PR #369, opened 2023-01-24
Show more than 10 results from goto cache

schinagl:
Currently Winfile only shows 10 results from the Goto Cache search after CTRL-G

This is many times not enough, so increased it

### DIFF
--- a/src/wfgoto.cpp
+++ b/src/wfgoto.cpp
@@ -518,29 +518,44 @@ VOID UpdateGotoList(HWND hDlg)
 	vector<PDNODE> options = GetDirectoryOptionsFromText(szText, &bLimited);
 
 	HWND hwndLB = GetDlgItem(hDlg, IDD_GOTOLIST);
+
+	// Since we're probably about to add a lot of items, suppress redraw
+	// and do it once at the end
+	SendMessage(hwndLB, WM_SETREDRAW, FALSE, 0);
 	SendMessage(hwndLB, LB_RESETCONTENT, 0, 0);
 
-	if (options.empty())
-		return;
-
-	constexpr int maxResults = 500;
-	for (size_t i = 0; i < maxResults && i < options.size(); i++)
+	if (!options.empty())
 	{
-		GetTreePath(options.at(i), szText);
+		const size_t resultCount = min(1000, options.size());
 
-		SendMessage(hwndLB, LB_ADDSTRING, 0, (LPARAM)szText);
+		// Try to allocate enough space in the list to avoid repeated
+		// allocations.  This is an optimization and doesn't need to be
+		// perfect, although being slightly too large is probably better
+		// than slightly too small.  Since building the paths for this
+		// seems needlessly expensive, assume each entry will be MAX_PATH.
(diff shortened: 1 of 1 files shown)


malxau:
I tried this out (on my low-end Atom.)  The initial letter typed is going to match a huge number of things and immediately hits this maximum.  On this system, it takes around 2 seconds to populate 1,000 items into the list.  Goto can't handle the second letter until the first has finished updating the list.  If I keep typing a "common" phrase then it becomes 2 seconds per key press; if I type something unique there's a 2 second hang initially followed by something more responsive later.

There's probably some elegant but more complex approach here of waiting for the user to pause before populating a larger number, but with the current code, 1000 seems too large for my CPU.  (It might be fine on your CPU.)  There might also be some cruder optimizations like having a lower maximum for 2 chars and below on the assumption that these are going to match many things and the user is still typing.

schinagl:
A low end Atom is not a benchmark and Winfile sure suffers performance problem on this machine in other areas too.

I reduced this to 500.

malxau:
I think there's a miscommunication here.  I didn't use this device as a kind of torture test.  This is the same device I'd use for winfile development, as well as other projects; I used the device because it had a git enlistment with winfile so it was the fastest way I could try out the PR, and this was my first impression.  I haven't had a bad experience with winfile on this device, because winfile was designed for 8 MHz 286s - this system is still at least two orders of magnitude faster than what winfile was designed for.

No, it's not a benchmark.  In this case a benchmark may not be that illuminating - if inserting 1000 items into a list took 100x longer than 10 items, what does mean in terms of experience?

Here's what I'm seeing.  This is x64 release, with 500 list entries:
500entries

Same thing, 10 entries:
10entries

What these don't really show is in the 500 case, the system is busy clearing and rebuilding the list, so it can't accept keyboard input.  The reason the animation takes longer is because it's waiting for the process to render before it can accept the next character; my typing speed is approximately constant.  In the 10 case, every frame in this animation is fully rendered - in effect, the screen capture tool is operating as a profiler, and every frame in the 500 case the system is trying to render, every frame in the 10 case it's completed.

With that, I made two changes.  First, calling `WM_SETREDRAW` with `FALSE` before `LB_RESETCONTENT`, and calling it again with `TRUE` after `LB_SETCURSEL`, followed by `RedrawWindow`.  So instead of trying to repaint 500 times, it does it once.  Second, call `LB_INITSTORAGE` before `LB_ADDSTRING`; in this case, we know the exact number of entries we want to insert, so the list box doesn't need to reallocate each time.

With these two changes, and 500 entries:
500entries_setredraw2

Rendering still takes time, although these screen captures show it always rendered (since it can't repaint when it's emptied.)  More importantly though, the input throttling is gone, so the UX is responsive.  Rendering still takes time, but it takes less time than for me to press the next key.

malxau:
I cleaned up the change and pushed it as commit 47e88f7733b204b44b47ca0adf72b7fd8cacc468 .  Please take a look and tell me what you think.

schinagl:
If disabling redraw helps, this really looks very good. I am fine
I was so keen to take it from your commit and change the init of resultCount to one line.

PR #370, opened 2023-01-31
PageUp/Down navigation for CTRL-G dialog

schinagl:
- Added PageUp/PageDown for goto-dialog navigation. Needed because with many items in the listcontrol one wants to quickly go through the cache hits.

- Removed Cursor up/down key wrap around, because the top/end can reached with VK_HOME/VK_END, thus no need for wrap around


malxau:
Would it be better to just forward these to the list control rather than try to manually apply the changes to the list control?  See commit 1049ec4c94cfd252851a3dcbbee6ebc6bc1aa1d1 .

schinagl:
Sure, good point, that's a much more lean and mean approach.

PR #371, opened 2023-02-05
Make en-us goto dialog resizeable

malxau:
This is really a follow on from #122 , in a cut down form while preserving the "full" change in the original PR.

- This uses DWLP_USER for the dialog context.  I'm a little uncomfortable with this because there's no way to "allocate" this space, it's just assuming nothing else is using it.  But it appears to work.
- Removes the VerifyResources change, so now VerifyResources will complain.
- Scopes it down to the minimal possible thing, randomly picked to be the en-us goto dialog, since making that scale to larger numbers of matches appears to be a theme, so being able to maximize it can help it be usable with more matches.


schinagl:
Looks good to me.
Shouldn't we apply this to all languages?

malxau:
As you've seen, my personal approach is to have larger and more comprehensive changes, so I think it should apply to all languages.  It's done this way because Craig was advocating for the smallest initial commit, which I think is something like this.  That has consequences (VerifyResources won't be happy.)

In theory, this code doesn't care:
1. For the most part, if the dialog style is not resizable, resize events will just never be sent.
2. This code uses hidden controls to determine placement of later controls.  If the controls aren't present, this code has nothing to do.
3. About the only "interesting" case is `RenderSizingGrip` and `IsWithinSizingGrip` where it looks at the window style to determine whether to display the grip; but since this is based on the window style, the code should work if invoked from a dialog which is not resizable according to its style.

schinagl:
I see. As far as I got Craigs concern in #122, it was mainly about storing RESIZE_DIALOG_INFO not stored with the window.
You now solved this by using DWLP_USER, which sounds a good solution to me.

I think the minimal *complete* solution contains all languages:
- If it works, we are fine, no nextra PR needed
- We have more people to test this
- If we fail it is the same effort to revoke with one language or many

So ;-) .... Be brave, lets change the languages too.

schinagl (approved):

PR #375, opened 2023-02-21
Remove unused defines

malxau:
This is to remove `#define` values that were carried forward from the Windows build environment and are not used.  These have mostly already been removed from C code but were still used in RC compilation.  I verified they are not used in winfile or in the SDK headers that it includes.

Note this changes `_M_IX86` to `_M_IX64` for x64 builds.  These should normally be defined by the compiler but are explicitly defined here.  It's not safe to get these wrong since it may result in incorrect structure definitions.  Since this is specific to the resource compiler it doesn't seem like a big risk, but still.


schinagl (approved):

schinagl:
Works on my side so at least a second pair of eyes is fine.
I think we are good to go

Issue #373, opened 2023-02-12
Open Explorer.exe in selected directory


schinagl:
Is it of general use to open Explorer.exe in the selected directory from Winfile?

I use Winfile all day and it is my 'base' for most operations. Thus I created the hotkey Ctrl-E to open an explorer .exe in the selected directory, much like the way as CTRL-K opens a command prompt.

Having an explorer opened via CTRL-E enables me to use a few right click shell extension very fast. So it is an improvement to me.

Should I push my current private PR? What do you think? Is it of general interest?

malxau:
Just my personal opinion (which isn't more valid than anyone else's)...

1. I think it'd be interesting to collect the situations where it's useful to switch from winfile to explorer, because those seem to be opportunities to improve winfile.
2. The current approach of hardcoding programs to open in a directory isn't going to work for everyone.

There's discussion in #44 about this, because as soon as it's possible to launch Bash, the next question is "which bash?"  Powershell attempts to address this by enumerating the registry and choosing the one with the largest index, assuming the user would always want the newest that they have installed.  Cmd seems completely overloaded with its ConEmu logic.  Part of my frustration with this is I have my own command shell (Yori), and it seems inappropriate to try to add support for that into Winfile in code for everyone, but launching it would be more useful to me than any of the current three.

If we wanted this to be extensible, I think any custom entry becomes:

- An open string.  I'd personally try to use the same model as Explorer for this, where there's some arbitrary command with a "%1" to expand with the requested directory.
  - The open string should be able to contain environment variables, so it can default to entries including %SystemRoot%, allowing Cmd and Powershell to be located without assuming the user's install location.
  - The "%1" can be optional since ExecProgram will try to launch anything with a specific current directory, which is frequently sufficient.
- An optional elevated open string, since Cmd requires something different for this.
- A shortcut key.  Maintaining compatibility with current behavior implies this can be almost anything, which is unfortunate but still achievable.

Part of me also would like to proactively parse these strings, check if the executable can be found, and _hide_ entries for executables that aren't found.  Bash and Powershell are both optional, don't exist on older versions, etc.

I'll also go on the record and assert that anyone who's knowledgeable enough to install ConEmu and use it with winfile is knowledgeable enough to configure winfile to use it, and its presence here is purely because there's no way to customize current behavior.  That said, the hiding logic above would make it possible to have a default entry for ConEmu that most users would never see.

[...]

PR #376, opened 2023-02-21
Open explorer

schinagl:
Little add-on which opens an explorer in the selected directory via File.Start E&xplorer menu or CTRL+E.
Is useful if your task can only be completed by e.g. Shell Extensions like archivers, renamers, ... from Explorer.exe

Has been discussed in #373


schinagl:
rebased onto master

src/wfcomman.c
```diff
@@ -1075,6 +1075,28 @@ AppCommandProc(register DWORD id)
...
+           StripFilespec(szDir);
+
+       if (GetSystemDirectory(szToRun, MAXPATHLEN) != 0)
+           lstrcat(szToRun, TEXT("\\..\\explorer.exe"));
+       else
+           lstrcpy(szToRun, TEXT("explorer.exe"));
+
+       TCHAR szParams[MAXPATHLEN] = { TEXT('\0') };
+
+       ret = ExecProgram(szToRun, szDir, szDir, FALSE, GetKeyState(VK_SHIFT) < 0);
```
malxau, line 1097: It doesn't matter, but what's the scenario for shift support? Normally there's only one explorer process, so elevating here will prompt the user but end up launching a window in the context of the unelevated explorer process (it will be functionally identical.)

schinagl: Good point, Will remove it

malxau (approved):

PR #377, opened 2023-02-25
Fix translation mistakes of zh introduced in #352

clzls:
Replace some misused `zh-Hant` translations with `zh-CN` ones. Mentioned in #352 earlier but some bugs leaked into the [final commit](https://github.com/microsoft/winfile/commit/4a689ad).

In short, a simple replacement: `符號鏈接`→`符号链接` (Symbolic link), `硬鏈接`→`硬链接` (Hard link), `目录连接点`→`目录联接` ((Directory) junctions) (from Windows cmdlet `mklink`), and restructuring of some sentences.


malxau-msft:
Thanks.  I double-checked that these are the strings used by mklink and that mechanical translation of the strings seems good.  (Also, the reverse translation of "directory join" seems so much nicer than the official English term of "directory junction".)

PR #379, opened 2023-03-08
OleDropSource Support

schinagl:
This is work in progress of OleDrop Source. Very work in progress ;-)
I guess the point of this branch currently is that the C implementation of the OleDropSource is there. At least I think that it should work.
I guess it is not working, because we put too little formats on the ole-drop source currently.

Have to investigate on how to do this, since this is rather undocumented. At least I believe


schinagl:
I am leaving Winfile

PR #392, opened 2023-05-25
Add 20px and 24px icons

malxau:
This adds 20x20 and 24x24 pixel icons.

Windows uses two system defined icon sizes, as returned by `GetSystemMetrics`.  These can change based on the DPI configuration of the monitor.  For more information on this, see https://learn.microsoft.com/en-us/windows/win32/uxguide/vis-icons#size-requirements .

If an icon with the specified size doesn't exist, Windows scales down larger sized icons to fit (as opposed to scaling up smaller ones.)  For icons consisting of high color gradients, this is generally fine.  For older icons with lower colors and lines, the result can be quite bad.  This PR is after seeing how the icon emerged on a 125% scaled system: http://www.malsmith.net/yori/yui_servercore.png

In theory, this should not change any other icon except adding these two sizes.  Unfortunately I don't think it'll be easy or obvious to see that in review.


schinagl (approved):
Opened the .ico file and looks fine.

PR #406, opened 2023-10-22
Tree pane updates when SymlinkDirs are created

schinagl:
When creating symlink dirs, the left tree pane didn' get updated.
This was caused by sending the wrong FSC code.

### DIFF
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -585,7 +585,7 @@ WFSymbolicLink(LPTSTR pszFrom, LPTSTR pszTo, DWORD dwFlags)
    Notify(hdlgProgress, IDS_COPYINGMSG, pszFrom, pszTo);
 
    if (CreateSymbolicLink(pszTo, pszFrom, dwFlags | (bDeveloperModeAvailable ? SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE : 0))) {
-      ChangeFileSystem(FSC_CREATE, pszTo, NULL);
+      ChangeFileSystem(dwFlags == SYMBOLIC_LINK_FLAG_DIRECTORY ? FSC_MKDIR : FSC_CREATE, pszTo, NULL);
       dwRet = ERROR_SUCCESS;
    } else {
       dwRet = GetLastError();


malxau-msft:
Sorry for the delay, I just got to look at this now.

Some observations:

1. The user can hide junctions.  This change will always update the left pane with a newly created junction even if hiding is enabled.
2. Similar to the above, a newly created junction in Winfile inherits the file attributes of the destination.  This means a new junction can be hidden/system, and will be displayed if junction hiding is off but hidden/system hiding is on.

I think the previous `FSC_MKDIR` code didn't need to think about this, because I can't find a way to create directories that are hidden/system, and the user can't hide directories.

Unrelated to this change, but related to the scenario and found trying things out:

3. File Manager uses `FindFirstChangeNotification` to monitor for file system changes.  Directory (and junction) creations outside of Winfile update the right pane but not the left pane.
4. It only monitors for changes within the directory that's opened, so has no way to be informed of a directory change that would be displayed in the left pane but is not within the opened directory.  This type of monitoring can be expensive so there's a good argument to not do it.  I think `FILE_NOTIFY_CHANGE_DIR_NAME` exists to facilitate this type of UI, so a program can be informed of any directory change on a volume without being informed of every file change on a volume.

schinagl:
Reverted tree pane update for junctions.
This will be a spereate PR

schinagl:
Since junction display was split of from this PR, this should be an easy one now....

schinagl:
> Sorry for the delay. I've been looking into this, and the symlink one, and both still have the issue where if the user is hiding hidden and system files, links get included in the left pane.
> 

Please do not mix up things!

Symbolic Links Dirs are shown regardless of any View/ByFileType setting of 'Hidden or System File/Junctions'
So your above quoted comment is wrong.

I intentionally split off an extra PR for Junction behaviour (#409) 

This PR only
* contains changes to send proper notification to the left pane when **Symbolic Link Dirs** are created, 
* contains one little obvious change. Please see the source changes of this PR

So you can approve this one.

schinagl:
I am leaving this Winfile Repo, no need to merge

PR #409, opened 2023-11-02
Tree pane updates when Junctions are created

schinagl:
Refresh left pane aka tree pane when junctions are created, but don't show them, if one has deselected junction appearance under View/File Type

### DIFF
--- a/src/lfn.c
+++ b/src/lfn.c
@@ -737,6 +737,7 @@ DWORD WFJunction(LPCWSTR pszLinkDirectory, LPCWSTR pszLinkTarget)
    }
 
    CloseHandle(hFile);
+   ChangeFileSystem(FSC_MKDIR, pszLinkDirectory, NULL);
    return ERROR_SUCCESS;
 }
 
--- a/src/treectl.c
+++ b/src/treectl.c
@@ -509,14 +509,26 @@ InsertDirectory(
        pNode->dwAttribs = dwAttribs;
    }
 
-   SendMessage(hwndLB, LB_INSERTSTRING, iMax, (LPARAM)pNode);
+   BOOL bInsertNode = TRUE;
+   if (pNode->dwAttribs & ATTR_JUNCTION)
+      if (!(GetWindowLongPtr(GetParent(hwndTreeCtl), GWL_ATTRIBS) & ATTR_JUNCTION))
+         bInsertNode = FALSE;
 
-   if (ppNode)
+   if (bInsertNode)
    {
-      *ppNode = pNode;
-   }
+      SendMessage(hwndLB, LB_INSERTSTRING, iMax, (LPARAM)pNode);
 
-   return (iMax);
+      if (ppNode)
+      {
+         *ppNode = pNode;
+      }
+      return iMax;
+   }
+   else 
+   {
+      LocalFree(pNode);
+      return 0;
+   }
 }
 
 
@@ -1188,7 +1200,7 @@ FillTreeListbox(HWND hwndTC,
                                &pNode,
                                IsCasePreservedDrive(DRIVEID(szDefaultDir)),
                                bPartialSort,
-                               (DWORD)(-1) );
+                               INVALID_FILE_ATTRIBUTES );
 
       if (pNode) {
 
@@ -2101,7 +2113,6 @@ TreeControlWndProc(
    INT    iSel;
    INT    i, j;
    INT    nIndex;
-   DWORD  dwTemp;
    PDNODE pNode, pNodeNext;
    HWND  hwndLB;
    HWND  hwndParent;
@@ -2466,14 +2477,14 @@ TreeControlWndProc(
          //
          // Insert it into the tree listbox
          //
-         dwTemp = InsertDirectory( hwnd,
-                                   pNode,
-                                   (WORD)nIndex,
-                                   szPath,
-                                   &pNodeT,
-                                   IsCasePreservedDrive(DRIVEID(((LPTSTR)lParam))),
-                                   FALSE,
-                                   (DWORD)(-1) );
+         InsertDirectory( hwnd,
+                          pNode,
+                          (WORD)nIndex,
+                          szPath,
+                          &pNodeT,
+                          IsCasePreservedDrive(DRIVEID(((LPTSTR)lParam))),
+                          FALSE,
+                          INVALID_FILE_ATTRIBUTES );
 
          //
          // Add a plus if necessary
--- a/src/treectl.c
+++ b/src/treectl.c
@@ -527,7 +527,10 @@ InsertDirectory(
    else 
    {
       LocalFree(pNode);
-      return 0;
+      if (ppNode)
+        *ppNode = NULL;
+
+        return 0;
    }
 }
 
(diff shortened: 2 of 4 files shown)


malxau-msft:
Sorry for the delay.  I've been looking into this, and the symlink one, and both still have the issue where if the user is hiding hidden and system files, links get included in the left pane.

Looking further into this, part of the problem is `WFMoveCopyDriverThread` calls `WFSymbolicLink` or `WFJunction`, then calls `WFSetAttr` to apply attributes from the destination to the link.  This is after `ChangeFileSystem` is called, so it doesn't have the opportunity to perform any hiding.  The `WFSetAttr` behavior seemed a little strange to me, since it's not what the API will do; but since it was code you added when adding symlink/junction support, I assume this is something you want to keep.

As I see it, that means either moving the `WFSetAttr` into `WFSymbolicLink`/`WFJunction`, or moving the `ChangeFileSystem` out of those and into `WFMoveCopyDriverThread` after it calls `WFSetAttr`.  Given that there are other similar functions that already call `ChangeFileSystem`, I'm assuming it's more consistent to move the call to `WFSetAttr`.

From there, `InsertDirectory` can check for `ATTR_HS`.  Unfortunately it also needs to `NULL` the output pointer (otherwise it will return freed memory) and the callers don't seem very diligent about checking for failure.  `ScanDirLevel` happens to be implementing the same checks as `InsertDirectory`, so if they're in perfect sync it won't touch the output pointer, but I think `ReadDirLevel` still can.  At least, I watched it crash accessing a freed pointer before setting it to `NULL`.

One odd quirk I found while looking into this is the root directory on NTFS has the hidden/system attributes set.  Previously this didn't matter since it wouldn't be returned from file enumeration so would never be hidden, but it's important to not hide it in `InsertDirectory` either.  If it is hidden, the tree will have zero items, causing the program to enter an infinite loop trying to populate the tree forever (where `InitFileManager` calls `TC_SETDRIVE`.)

The changes I've been working with are at https://github.com/malxau-msft/winfile/commit/485bd11cb96681290780f893694b5055afb7038d .  I'm not attached to anything here and am not very confident in the error handling from `InsertDirectory`, although it appears to work for all of the cases I've tried so far.

schinagl:
> Sorry for the delay. I've been looking into this, and the symlink one, and both still have the issue where if the user is hiding hidden and system files, links get included in the left pane.
> 
You are mixing up things I gess. See my comment with #406. 
> 
> Looking further into this, part of the problem is `WFMoveCopyDriverThread` calls `WFSymbolicLink` or `WFJunction`, then calls `WFSetAttr` to apply attributes from the destination to the link. This is after `ChangeFileSystem` is called, so it doesn't have the opportunity to perform any hiding. The `WFSetAttr` behavior seemed a little strange to me, since it's not what the API will do; but since it was code you added when adding symlink/junction support, I assume this is something you want to keep.
> 
> As I see it, that means either moving the `WFSetAttr` into `WFSymbolicLink`/`WFJunction`, or moving the `ChangeFileSystem` out of those and into `WFMoveCopyDriverThread` after it calls `WFSetAttr`. Given that there are other similar functions that already call `ChangeFileSystem`, I'm assuming it's more consistent to move the call to `WFSetAttr`.

Completley disagree! We must not mix up changing creation of symbolic links and changing attributes.
This is bad design and no other function in Winfile does this.

> 
> From there, `InsertDirectory` can check for `ATTR_HS`. Unfortunately it also needs to `NULL` the output pointer (otherwise it will return freed memory) and the callers don't seem very diligent about checking for failure. `ScanDirLevel` happens to be implementing the same checks as `InsertDirectory`, so if they're in perfect sync it won't touch the output pointer, but I think `ReadDirLevel` still can. At least, I watched it crash accessing a freed pointer before setting it to `NULL`.
> 

I only agree with measures on setting ppNode to nullptr as shown in your snippet, but a check for `ATTR_HS seems not related to this PR. Which problem are you addressing? 

> One odd quirk I found while looking into this is the root directory on NTFS has the hidden/system attributes set. Previously this didn't matter since it wouldn't be returned from file enumeration so would never be hidden, but it's important to not hide it in `InsertDirectory` either. If it is hidden, the tree will have zero items, causing the program to enter an infinite loop trying to populate the tree forever (where `InitFileManager` calls `TC_SETDRIVE`.)

This was never a problem before, why should it be a problem now?
Please describe a use-case of failure.

**Summary**
Can you please tell me the problem with my solution. (except ppNode = nullptr, which I am about to add. Thx! )
To me it seems you are describing other problems not related to this PR with ATTR_HS & co

So please describe a use-case where my solutions fails.

malxau-msft:
> So please describe a use-case where my solutions fails.

1. Under View Options, ensure "Show Junction Points" is set, and "Show Hidden/System Files" is **not** set.
2. Create a directory.
3. Mark that directory hidden.
4. Create a junction that points to that directory.
5. Restart winfile.

Without this PR:

1. The newly created junction will not be displayed in the left pane or the right pane.
2. On restart, the junction will not be displayed in the left pane or the right pane.

With this PR:

1. The newly created junction will be displayed in the left pane but not the right pane.
2. On restart, the junction will not be displayed in the left pane or the right pane.

Normally, Winfile applies `ATTR_HS` filtration in `WFFindFirst` and `WFFindNext`.  This happens before the reparse tag is inspected, and applies to symbolic links in the same way as junctions.  Calling `InsertDirectory` directly bypasses this filtration, so this PR is populating the left pane with things that it normally wouldn't contain.

> Completley disagree! We must not mix up changing creation of symbolic links and changing attributes.
This is bad design and no other function in Winfile does this.

Feel free to propose an alternative solution.  Looking today, another valid approach would be to ensure `WFSetAttr` refreshes the left pane (it's already calling `ChangeFileSystem`.)

malxau-msft:
This is the behavior I am seeing:

https://github.com/microsoft/winfile/assets/104935627/e79d3378-5297-4c2a-ad17-d1658f928182

>This problem is, that a freshly hidden directory does not vanish on the spot when you set it to hidden

Yes, I agree.  The issue is that Winfile will set these junctions to hidden at the time they are created (at https://github.com/microsoft/winfile/blob/7ad683989e178a8fe6f5f9abfecb9821d0b56958/src/wfcopy.c#L2822), so it creates this condition in a way that is hidden from the user.  That's why I'm suggesting `WFSetAttr` would be one way to fix it - it fixes the general case, which also fixes the case of a freshly created junction to a hidden directory.

schinagl:
My usecase up till now was: create a junction inside a hidden directory. This works correctly
https://github.com/microsoft/winfile/assets/7418603/916d8bd1-57a1-4d18-9b9f-06d1ce8bd08a
ok. So far so good

Your use-case was
Create a junction to a hidden directory, which I can reproduce, but I see no difference in behaviour with or without this PR

Without this PR
https://github.com/microsoft/winfile/assets/7418603/8e154059-b0c2-4ee8-83f9-5a40cc17bbec

With this PR
https://github.com/microsoft/winfile/assets/7418603/cbfe8866-0418-420e-ab14-877d27f9c101

One can not see the junction in both cases, but that's obvious, because the unchecked 'Show Hidden' has precedence over 'Show Junctions'

So what's the problem?

schinagl:
I am leaving this Winfile Repo, no need to merge

PR #410, opened 2023-11-05
allow right clicking multiple files

MamiyaOtaru:
probably a better way to do this.  Basically checks if the right clicked item is one of the already selected items and if so doesn't change the selection.  When an action is carried out, the full list is then returned to be acted on

### DIFF
--- a/src/wfdir.c
+++ b/src/wfdir.c
@@ -886,6 +886,29 @@ DirWndProc(
       ActivateCommonContextMenu(hwnd, hwndLB, lParam);
          break;
 
+   case FS_RCLICKED:
+   {
+       LPXDTALINK lpxdta;
+       DWORD item = wParam;
+       INT iMac = (INT)SendMessage(hwndLB, LB_GETSELCOUNT, 0, 0L);
+       LPINT lpSelItems = (LPINT)LocalAlloc(LMEM_FIXED, sizeof(INT) * iMac);
+       iMac = (INT)SendMessage(hwndLB,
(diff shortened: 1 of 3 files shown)


MamiyaOtaru:
intended to address  [Right Mouse Click Context Menu issues #260 ](https://github.com/microsoft/winfile/issues/260)

malxau-msft:
Thanks @MamiyaOtaru ! Unfortunately the CLA is still showing as not being signed, and we can't move forward with this until it is.

Is there a reason for not using `LB_GETSEL`?  As far as I can tell, this would eliminate the need for the allocation and loop.

MamiyaOtaru:
@microsoft-github-policy-service agree


"Is there a reason for not using LB_GETSEL?" - lack of familiarity with the code.

malxau-msft:
Thanks for taking care of the CLA.  What I was trying to say about `LB_GETSEL` is in commit e14a3ca32a363f7ad1f7b29c8ac0fb3e01164103 - I think this is doing the same thing?  Is there something I missed?

Also while looking at this again today, the code immediately below this is checking if the first selected item is a directory, and if so, disable the edit item.  Unfortunately both open and edit don't work for multiple files - see the code in `OpenOrEditSelection` which calls `FS_GETSELECTION`.  6ceaee602d3bc35bbfe104484330035cba8b9106 disables open/edit if multiple objects are selected, which seems like the easiest path.  A better change would be to change `OpenOrEditSelection` to handle this, but it seems distinct from the multiple selection change.

PR #417, opened 2024-02-23
Update winfile version

malxau-msft:
Seems like a good time to prepare a new release.

This is my first time doing this, so apologies in advance for any bumps, particularly with Store and Chocolatey releases.  Getting signed binaries on Github looks straightforward.

### DIFF
--- a/Package/Package.appxmanifest
+++ b/Package/Package.appxmanifest
@@ -7,7 +7,7 @@
   xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10"
   xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
   IgnorableNamespaces="desktop uap uap3 mp rescap">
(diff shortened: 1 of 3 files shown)


schinagl:
There is a flaw in your .sln and projects
In one of the .vcxprojs, you have the Any CPU configuration, but not in all.
If you open a .sln via IDE, which does not have 'Any CPU' and there is one .vcxproj with 'Any CPU', all .sln will get 'Any CPU' and the .sln is modified by the IDE
Since you you don't use the IDE you maybe missed this

malxau-msft:
> If you open a .sln via IDE, which does not have 'Any CPU' and there is one .vcxproj with 'Any CPU', all .sln will get 'Any CPU' and the .sln is modified by the IDE Since you you don't use the IDE you maybe missed this.

@schinagl I created #430 for this.  Please let me know if it addresses this concern.

schinagl:
#430 fixes the 'Any CPU' issue

PR #429, opened 2024-03-17
Delete stray Windows icon

wesinator:
unused icon


malxau-msft:
Heh, this is a bit of a puzzle.  Unlike every other change, I don't see the commit that introduced this file (because it's binary?)

Here's my theory: Windows 3.0 did not have `ShellAbout`, and the WinFile about dialog used the Windows icon.  The 3.1 migration to `ShellAbout` allowed this to be removed from the resource, but it's still in version control through NT and 95.  Winver seems to have done the same thing; its icon was removed in the Vista timeframe, but Winfile was removed completely before then.

malxau-msft (approved):

clzls:
> Unlike every other change, I don't see the commit that introduced this file (because it's binary?)

For this repo, it was introduced in commit [068919f](https://github.com/microsoft/winfile/commit/068919f61dbc35cbf656ab14ebf19fceae68205e), the tree's root. I wonder when it came into the source tree before this GitHub repo...

malxau-msft:
> I wonder when it came into the source tree before this GitHub repo...

What I'm trying to say is it's not exactly clear.  The revision control system used in early NT tracks the final state of files, as well as a chain of differences.  Here, I see the icon in the final state, but it is never recorded in the difference chain, which means I can't track the commit introducing it.

However, what I can see is the .rc file has never referenced it, back to the beginning of the NT port of this code in April 1991.  And I can see the file in the Windows 3.1 tree, also without being referenced (although I don't have differences for that, just a snapshot.)  I do not have a source snapshot of 3.0, so am guessing a little about that.

The April 1991 NT port also contains `ShellAbout` (it never had the 3.0 hand rolled about dialogs.)  Winfile would have needed an icon to display in the hand rolled about dialog when it was part of the binary.  Note though the icon here is a 3.1 era icon, not the 3.0 era icon, so for this theory to be true, somebody updated this icon early in 3.1 development, then the dialogs became shared, and nobody went back to delete the icon.

Issue #432, opened 2024-03-20
Wanted: Undelete DLL


malxau-msft:
Winfile has a facility to load a third party undelete DLL that enables a menu item to invoke it.

I haven't been able to find any 32 bit implementation of this interface, so am asking if anyone else using Winfile knows of one.

For 16 bit, Norton Desktop implemented it (https://jeffpar.github.io/kbarchive/kb/085/Q85315/) but I don't see any 32 bit version of that product.

PR #433, opened 2024-03-23
Remove undelete which seems unused

malxau-msft:
Following on from #432 , this PR removes Undelete support, which I'm fairly sure is unused (and possibly unusable.)

In DOS, Undelete tools were possible because deleting a file just marks a flag in the directory entry, and DOS had no effective file system metadata cache, so a tool needed to scan a directory looking for deleted file, clear a flag, and DOS would start using it.

On a modern system, file systems have very extensive in memory caching and there's no way to bring a file back from the dead behind its back.  About the "best" way I can imagine is to scan the MFT looking for deleted file records, see if they still refer to clusters, read those blocks through the partition (not through the file system) and write them into a new file.  But given the way NTFS recycles file records, this is an exceedingly temporary recovery window.

I don't think it's a coincidence that moving to modern 32 bit caching file systems caused the Recycle Bin to exist, allowing file recovery to take place from above the file system layer rather than underneath it.  There are NT4 era Undelete products, which often include a driver to monitor deletions (ie., a kernel mode recycle bin.)  I'm yet to find one that targeted NT 3.x with Winfile integration.


schinagl:
Remove it. It *is* obsolete. 

But: one can/could undelete on NTFS via sysinternals undelete. But it is history.

PR #438, opened 2024-03-31
Enable SdlCheck and fix warning for use of uninitialized pointer value

malxau-msft:
As part of generating an official build, one warning was issued for not using SdlCheck (see https://learn.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks?view=msvc-170 .)  Enabling this has the effect of enabling use of uninitialized pointer local variables, and most of this change revolves around that.  The general pattern seems to be Winfile using two locals, one which is initialized if a second is set, then later consumed if the second is set.  The compiler is not smart enough to understand that whenever it is used, the local will be initialized.  The changes here have the effect of suppressing the compiler warning, although any use of the local where it hasn't been initialized with the correct value will still be incorrect.  About the best thing these changes do is force a deterministic failure rather than consume uninitialized stack and fail in random ways.

This also splits the XP build definition from the regular one, since it seems to have enough XP-specific switches now that doing so is cleaner.  SdlCheck is only enabled for the 2022 compiler, so it's not present in the XP build.

### DIFF
--- a/src/treectl.c
+++ b/src/treectl.c
@@ -616,19 +616,19 @@ ReadDirLevel(
    LPTSTR  szAutoExpand,
    BOOL bPartialSort)
 {
-   LPWSTR      szEndPath;
+   LPWSTR    szEndPath;
    LFNDTA    lfndta;
    INT       iNode;
    BOOL      bFound;
-   PDNODE     pNode;
+   PDNODE    pNode;
    BOOL      bAutoExpand;
    BOOL      bResult = TRUE;
    DWORD     dwView;
    HWND      hwndParent;
    HWND      hwndDir;
    LPXDTALINK lpStart;
-   LPXDTA*  plpxdta;
-   LPXDTA   lpxdta;
+   LPXDTA*   plpxdta;
+   LPXDTA    lpxdta;
    INT       count;
 
    UINT      uYieldCount = 0;
@@ -649,6 +649,8 @@ ReadDirLevel(
    //
 
    lpStart = NULL;
+   plpxdta = NULL;
+   lpxdta = NULL;
 
    if (!(dwView & VIEW_PLUSES)) {
 
@@ -1034,6 +1036,8 @@ StealTreeData(
    DWORD dwView;
    DWORD dwAttribs;
 
+   hwndT = NULL;
+
    //
    // we need to match on these attributes as well as the name
    //
@@ -1159,6 +1163,7 @@ FillTreeListbox(HWND hwndTC,
    BOOL bPartialSort;
    DRIVE drive;
 
+   pNode = NULL;
 
    hwndLB = GetDlgItem(hwndTC, IDCW_TREELISTBOX);
 
@@ -1226,7 +1231,9 @@ FillTreeListbox(HWND hwndTC,
       FindItemFromPath(hwndLB, szDefaultDir, FALSE, NULL, &pNode);
    }
 
-   SendMessage(hwndLB, LB_SELECTSTRING, (WPARAM)-1, (LPARAM)pNode);
+   if (pNode != NULL) {
+      SendMessage(hwndLB, LB_SELECTSTRING, (WPARAM)-1, (LPARAM)pNode);
+   }
 
    UpdateStatus(GetParent(hwndTC));  // Redraw the Status Bar
 
@@ -3170,6 +3177,7 @@ SameSelection:
          BOOL bDir;
          BOOL bChangeDisplay = FALSE;
 
+         hwndNext = NULL;
          TypeAheadString('\0', NULL);
          GetTreeWindows(hwndParent, NULL, &hwndDir);
 
@@ -3207,7 +3215,7 @@ SameSelection:
                }
                else
                {
-                   hwndSet = hwnd;
+                  hwndSet = hwnd;
                }
             }
          }
--- a/src/wfassoc.c
+++ b/src/wfassoc.c
@@ -1783,6 +1783,9 @@ ClassesRead(HKEY hKey,
 
    iKey = 0;
 
+   pFileType = NULL;
+   pExt = NULL;
+
    pFileTypePrev = *ppFileTypeBase;
    pExtPrev = *ppExtBase;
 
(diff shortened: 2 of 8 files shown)


src/treectl.c
```diff
@@ -649,6 +649,8 @@ ReadDirLevel(
    //
 
    lpStart = NULL;
+   plpxdta = NULL;
```
schinagl, line 652: Pleae initialize with declaration in one line

src/treectl.c
```diff
@@ -649,6 +649,8 @@ ReadDirLevel(
    //
 
    lpStart = NULL;
```
schinagl, line 651: Pleae initialize with declaration in one line

src/treectl.c
```diff
@@ -649,6 +649,8 @@ ReadDirLevel(
    //
 
    lpStart = NULL;
+   plpxdta = NULL;
+   lpxdta = NULL;
```
schinagl, line 653: Please initialize with declaration in one line

src/treectl.c
schinagl, line 652: Pleae initialize with declaration in one line

src/treectl.c
```diff
@@ -1034,6 +1036,8 @@ StealTreeData(
    DWORD dwView;
    DWORD dwAttribs;
 
+   hwndT = NULL;
```
schinagl, line 1039: Pleae initialize with declaration in one line

src/treectl.c
```diff
@@ -1159,6 +1163,7 @@ FillTreeListbox(HWND hwndTC,
    BOOL bPartialSort;
    DRIVE drive;
 
+   pNode = NULL;
```
schinagl, line 1166: Pleae initialize with declaration in one line

src/treectl.c
```diff
@@ -3170,6 +3177,7 @@ TreeControlWndProc(
          BOOL bDir;
          BOOL bChangeDisplay = FALSE;
 
+         hwndNext = NULL;
```
schinagl, line 3180: Pleae initialize with declaration in one line

src/wfassoc.c
```diff
@@ -1783,6 +1783,9 @@ ClassesRead(HKEY hKey,
 
    iKey = 0;
 
+   pFileType = NULL;
```
schinagl, line 1786: Pleae initialize with declaration in one line

src/wfassoc.c
```diff
@@ -1783,6 +1783,9 @@ ClassesRead(HKEY hKey,
 
    iKey = 0;
 
+   pFileType = NULL;
+   pExt = NULL;
```
schinagl, line 1787: Pleae initialize with declaration in one line

src/wfcopy.c
```diff
@@ -1403,6 +1403,7 @@ GetNextPair(PCOPYROOT pcr, LPTSTR pFrom,
    STKCHK();
    *pFrom = CHAR_NULL;
    *pdwError = 0 ;
+   pDTA = NULL;
```
schinagl, line 1406: Please initialize with declaration in one line

src/wfdrives.c
```diff
@@ -749,6 +749,8 @@ DrivesWndProc(HWND hWnd, UINT wMsg, WPARAM wParam, LPARAM lParam)
                    BOOL bDir;
                    BOOL bChangeDisplay = FALSE;
 
+                   hwndNext = NULL;
```
schinagl, line 752: Please initialize with declaration in one line

src/wfinfo.c
```diff
@@ -1059,6 +1059,8 @@ UpdateDriveListWorker(VOID)
 
    INT iUpdatePhantom = iUpdateReal ^ 1;
 
+   hEnum = NULL;
```
schinagl, line 1062: Please initialize with declaration in one line

src/wfutil.c
```diff
@@ -1122,9 +1122,11 @@ SetDlgDirectory(HWND hDlg, LPTSTR pszPath)
...
   HANDLE    hFont;
   HANDLE    hFontBak;
-  TCHAR      szPath[MAXPATHLEN+5];
-  TCHAR      szTemp[MAXPATHLEN+20];
-  TCHAR      szMessage[MAXMESSAGELEN];
+  TCHAR     szPath[MAXPATHLEN+5];
+  TCHAR     szTemp[MAXPATHLEN+20];
+  TCHAR     szMessage[MAXMESSAGELEN];
+
+  hFontBak = NULL;
```
schinagl, line 1129: Please initialize with declaration in one line

src/wfdir.c
```diff
@@ -1318,6 +1318,8 @@ ChangeDisplay(
 
    case WM_CREATE:
 
+      lpStart = NULL;
```
schinagl, line 1321: Why does this satisfy the SdlCheck?
lpStart is used above many times

malxau-msft: The compiler understands codeflow, it just doesn't understand the relationship between multiple variables.  Here, the outer switch statement has three case branches.  The first two initialize `lpStart` on line 1064.  These also have several goto statements that can jump into the `WM_CREATE` case, but only after `lpStart` is initialized.  In the `WM_CREATE` case (without any goto) it's initialized because `bCreateDTABlock` must be `TRUE` - but that exceeds the compiler's reasoning capabilities.  Without doing something to cover `WM_CREATE`, the compiler complains at line 1478 because it doesn't know that it must have taken the `bCreateDTABlock` branch above.

schinagl: I see. Thx. 
To make the intention of this initialization more obvious to the human reviewer, please initialize it in line 1000 with the declaration, which then has the same effect

malxau-msft: The earlier these initializations occur, the less effective the warning is.  The warning can catch when values are used that are not initialized; it cannot catch values that are initialized with an invalid value such as NULL.  Keeping it here means that if any code above it tries to use the variable without initializing it, it can still be caught.  The frustrating part is most of the time this isn't possible, which limits the value of the warning.

schinagl:
It is about initialization and not about catching exception. This is the most weird argument I have ever heard. 

As so many times I agree to disagree with your programming style from the 80s and your attitude of handling changes and running this repo. 

This repo has detoriated into @malxau private environment.

Issue #100, opened 2018-04-15
'Goto Directory' ignores hidden folders


cgaerter:
I have set my (Windows 7) system to not display hidden files/folders, but since the path to the desktop starts with the hidden users folder in the system root, I cannot navigate to the desktop or to my personal files.

I suggest to either add an option to WinFile to display hidden files/folders without changing the system-wide setting or to enable the 'Goto Directory' feature to jump to folders below a hidden folder.

craigwi:
Hidden and other types of directories can certainly be indexed.  In my use case, doing so resulted in too much noise.  But I think an option makes sense. 

We probably need to start an options dialog for the goto command since another obvious fix to make is to index other drives if desired.

DBJDBJ:
IMHO including hidden folders will indeed make humongous amount of noise
other drives are it seems already indexed? let me check once more ...

Ok, I have C:, D: H: drives and their windows open , while on D: , if I do CTRL+G and type "u", "c:\users" is immediately shown ...

I close C: and H: windows and still same as before ..  I do CTRL+G and type "u", "c:\users" is immediately shown ...

I assume this was fixed before I joined the project :)

DBJDBJ:
IMHO the "obvious fix" might be to search only the active drive ... 
right now wherever, on the disks, one is, only C: is indexed (true) and if user desired to "go to" the result current window changes to c: but the previously active drive stays as selected on the drives tool bar. 

I assume this would be easy to fix ...

HTH

toiletflusher:
Old issue post/thread, i know, but you can enable hidden files and folders from winfile already. Even in the NT 4.0 version.
Click the View menu, Click on by file type...
Then it lets you choose to show it or not. it also fixes the problem that Goto Directory ignores hidden folders.

JBensimon99:
Yes, showing hidden files and folders is an available option, BUT that setting is not preserved when saving settings for future launches of WinFile (which by the way are stored in "%AppData%\Roaming\Microsoft\Winfile\WINFILE.INI"). This should be an easy fix.

malxau-msft:
I think everyone here is describing a slightly different thing.

- @JBensimon99 Can you describe more about the sequence of steps that cause the hidden display of files to not be saved? I verified that this is saved for me on application exit as well as via explicit save.
[...]

PR #446, opened 2024-07-15
Have a hidden option for goto to cache hidden and system subtrees

malxau-msft:
This is to address the original issue in #100 .  In the previous code (and after this, still default code), hidden and system directories are not enumerated.  This means any subdirectories under a hidden/system directory will not be cached.

This PR adds a UI-less `winfile.ini` option to include hidden and system directories, including their children.


Sheikhpk (approved):

schinagl:
Even if it is merged, one note:
This PR also contains a change towards C89 even if wfgoto.cpp is a CPP file. Sic. 
Please get used it to that C89 *is* outdated, and join us in a modern world with respecting the [principle of locality](https://medium.com/@prabh27/the-principle-of-locality-fcd0788c2667)

PR #457, opened 2025-02-22
Add brazilian portuguese translation

AFaustini:
Sending the translation to Brazilian Portuguese, using the pt-PT language files as a base, only changing the few words that differ between the sub-languages.


AFaustini:
@microsoft-github-policy-service agree

src/lang/res_pt-BR.rc
```diff
@@ -0,0 +1,644 @@
...
+BEGIN
+    POPUP       "&Arquivo"
+    BEGIN
+    MENUITEM    "&Abrir\tEnter",     IDM_OPEN
+    MENUITEM    "Editar\tF12",        IDM_EDIT
+    MENUITEM    "Mo&ver...\tF7",     IDM_MOVE
+    MENUITEM    "&Copiar...\tF8",     IDM_COPY
+    MENUITEM    "S&ymlink...\tF11", IDM_SYMLINK
+    MENUITEM    "Hardlin&k...\tShift+F11", IDM_HARDLINK
+    MENUITEM    "Copiar para a Clip&board\tCtrl+C", IDM_COPYTOarea de transferência
```
malxau-msft: Looks like an automatic search/replace messed up the ID here

src/lang/res_pt-BR.rc
```diff
@@ -0,0 +1,644 @@
...
+    MENUITEM    "Cortar para a area de transferência\tCtrl+X", IDM_CUTTOarea de transferência
+    MENUITEM    "&Colar\tCtrl+V",   IDM_PASTE
+    MENUITEM    "Ex&cluir...\tDel",  IDM_DELETE
```
malxau-msft: Note this is changing the accelerator key.  The menu now has three `c`s.  Windows handles this by having the user press `c` repeatedly to select them, then pressing Enter to activate, but the experience is messy and I try to avoid overloading like this.

Are you sure using the same key for multiple items is the right thing to do?

src/lang/res_pt-BR.rc
```diff
@@ -0,0 +1,644 @@
...
+    POPUP       "Ár&vore"
+    BEGIN
+    MENUITEM    "&Expandir um Nível\t+", IDM_EXPONE
+    MENUITEM    "Expandir &Ramo\t*",    IDM_EXPSUB
+    MENUITEM    "Expandir &Tudo\tCtrl+*",  IDM_EXPALL
+    MENUITEM    "&Fechar Ramo\t-",  IDM_COLLAPSE
+    MENUITEM    SEPARATOR
+    MENUITEM    "&Indicar Ramos Expansíveis",    IDM_ADDPLUSES
+    END
+    POPUP       "V&er"
```
malxau-msft: The original used "Exibir" here - is that better?

src/lang/res_pt-BR.rc
```diff
@@ -0,0 +1,644 @@
...
+    BEGIN
+    MENUITEM    "Á&rvore e Diretório",  IDM_BOTH
+    MENUITEM    "&Apenas Árvore",           IDM_TREEONLY
+    MENUITEM    "Apenas &Diretório",      IDM_DIRONLY
+    MENUITEM    SEPARATOR
+    MENUITEM    "D&ividir",               IDM_SPLIT
+    MENUITEM    SEPARATOR
+    MENUITEM    "&Nome",                IDM_VNAME
+    MENUITEM    "&Todos os Detalhes dos Arquivos",    IDM_VDETAILS,   CHECKED
```
malxau-msft: The original used "do Arquivo" here - is that better?

src/lang/res_pt-BR.rc
```diff
@@ -0,0 +1,644 @@
...
+    MENUITEM    "&Personalizar a Barra de Ferramentas...", IDM_TOOLBARCUST
+    MENUITEM    "Preferências &Adicionais", IDM_PREF
+    MENUITEM    SEPARATOR
+    MENUITEM    "&Barra de Ferramentas",                 IDM_TOOLBAR
+    MENUITEM    "Barra de &Unidades",                IDM_DRIVEBAR
+    MENUITEM    "Barra de &Estado",              IDM_STATUSBAR
+    MENUITEM    SEPARATOR
+    MENUITEM    "Executar &Minimizado",     IDM_MINONRUN
+    MENUITEM    "Criar &Index 'Ir para' ao Ligar", IDM_INDEXONLAUNCH
+    MENUITEM    "&Salvar Definições ao Sair",   IDM_SAVESETTINGS
```
malxau-msft: The original used "Configurações" here - is that better?

src/lang/res_pt-BR.rc
```diff
@@ -0,0 +1,644 @@
...
+
+// Help strings for the "standard" toolbar items.
+
+STRINGTABLE
+BEGIN
+    MH_MYITEMS+IDM_OPEN,        "Abre o item selecionado"
+    MH_MYITEMS+IDM_EDIT,        "Editar o item selecionado usando o notepad.exe"
+    MH_MYITEMS+IDM_MOVE,        "Move o item selecionado"
+    MH_MYITEMS+IDM_COPY,        "Copia Arquivos e Diretórios"
+    MH_MYITEMS+IDM_COPYTOarea de transferência, "Copia um ou mais Arquivos para a area de transferência"
```
malxau-msft: This looks like another search/replace messing up the ID

src/lang/suggest_pt-BR.db
```diff
@@ -0,0 +1,90 @@
...
+SUGGEST( 2, DE_MANYSRC1DEST, 0L, \
+ "Seleccione apenas um arquivo para renomear, ou use wildcards (por examplo, *.TXT) para renomear um grupo de arquivos com nomes semelhantes." )
+SUGGEST( 3, DE_DIFFDIR, 0L, \
+ "Não é possível renomear para um diretório ou disco diferente.\nUse o comando Mover arquivo." )
+SUGGEST( 4, DE_ROOTDIR, 0L, \
+ "Não é possível renomear o diretório raíz." )
+SUGGEST( 5, DE_DESTSUBTREE, 0L, \
+ "O diretório de destino é um subdiretório da fonte." )
+SUGGEST( 6, DE_WINDOWSFILE, 0L, \
+ "arquivo em uso pelo Windows." )
```
malxau-msft: Is this the best capitalization?

src/lang/res_pt-BR.rc
```diff
@@ -0,0 +1,644 @@
...
+   Copyright (c) Microsoft Corporation. All rights reserved.
+   Licensed under the MIT License.
+
+********************************************************************/
+
+LANGUAGE LANG_PORTUGUESE, SUBLANG_PORTUGUESE_BRAZILIAN
+
+FRAMEMENU MENU
+BEGIN
+    POPUP       "&Arquivo"
```
malxau-msft, line 14: This is the same accelerator as used for `Ajuda`, making opening it with the keyboard a pain.  The original used `?` instead of `Ajuda` presumably for this reason.  Is there a common convention for representing these two, since both seem very common?

malxau-msft (changes requested):
The clipboard IDs need to be updated for this code to compile.

I'd suggest looking through the accelerator keys to confirm that they are the best choices.  Seeing the same key used for multiple menu items seems difficult to use.  Part of that was inherited from pt-PT, but this change seems to introduce more, in more common parts of the UI.

The translations themselves seem fine though.

AFaustini (commented):
I`ve made the requested changes

